Issue #121, opened 2015-01-23
Documentation issue on developer instructions


litlfred:
The documentation here
 http://rapidpro.github.io/rapidpro/docs/development/
could use a bit more explanation for those that are new to some of the tools.

For example this is what I had to do to get postgres configured with Peer authentication:
  sudo apt-get install postgresql-client  postgis*
  sudo -u  postgres createuser temba --pwprompt -d
  sudo adduser temba
  sudo -u temba  psql --user=temba postgres
    postgres=> create database temba;
    CREATE DATABASE
    \q

To get lessc and coffee-script installed:
   sudo apt-get install node node-less coffee-script
   

For building the virual environment
   sudo apt-get install python-virtualenv  postgresql-server-dev-9.3 python-dev ncurses-dev

Not sure what level of detail you want in the install instructions.

nicpottier:
Cool, can you give us a pull request? (all the docs are on the gh-pages branch)

litlfred:
Will try to (once I make sure everything is working).   in the meantime i will keep updating the orignial comment where there are gaps.

allyshaban5:
The added command "sudo adduser temba" is of help.
Thank you Carl.

litlfred:
See more here:
   https://github.com/rapidpro/rapidpro/issues/121

litlfred:
@nicpottier it would be helpful to put directions for changing the port.   can you give me a hint?

nicpottier:
Changing what port? RapidPro is a start Django app, so the same startup options are available:
  https://docs.djangoproject.com/en/1.7/ref/django-admin/#runserver-port-or-address-port

litlfred:
OK.  thanks - I haven't done any Django development, so didn't know.

litlfred:
OK to put a disclaimer that these directions ar for Ubuntu 14.04?

nicpottier:
We actually run on 12.04, but I can double check that things aren't too specific.

litlfred:
One last hint to get through the install (hopefully) if you don't mind -- getting the error below.  Any suggestions?

(env)litlfred@masala:~/rapidpro$ python manage.py syncdb
Traceback (most recent call last):
  File "manage.py", line 9, in <module>
    execute_from_command_line(sys.argv)
  File "/home/litlfred/rapidpro/env/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
[...]

PR #124, opened 2015-01-26
Fixes for developer install documentation

litlfred:
Fixes issues raised in thread here https://github.com/rapidpro/rapidpro/issues/121

### DIFF
--- a/_docs/development.md
+++ b/_docs/development.md
@@ -65,6 +65,8 @@ $ cd rapidpro/temba
 $ ln -s settings.py.dev settings.py
 {% endhighlight %}
 
+At this point, you should edit the rapidpro/settings.py to adjust the password for the postgresql connection.
+
 ##Install Node 
 {% highlight bash %}
 $ sudo apt-get install node npm coffee-script
--- a/_docs/development.md
+++ b/_docs/development.md
@@ -69,8 +69,7 @@ At this point, you should edit the rapidpro/settings.py to adjust the password f
 
 ##Install Node 
 {% highlight bash %}
-$ sudo apt-get install node npm coffee-script
-$ sudo npm install -g less
+$ sudo apt-get install node node-less coffee-script
 {% endhighlight %}
 
 ## Build virtual environment
--- a/_docs/development.md
+++ b/_docs/development.md
@@ -25,23 +25,26 @@ You'll need the following to get started:
 ## Create temba user for PostgreSQL
 
 {% highlight bash %}
-$ createuser temba --pwprompt -d
+$ sudo apt-get install postgresql-client postgresql-server-dev-9.3 postgis*
+$ sudo -u postgres createuser temba --pwprompt -d
 Enter password for new role:
 Enter it again:
+$ sudo adduser temba
 {% endhighlight %}
 
 ## Create temba database, add PostGIS
 
 Create the database as temba user:
 {% highlight bash %}
-$ psql --user=temba postgres
+$ sudo -u temba psql --user=temba postgres
 postgres=> create database temba;
 CREATE DATABASE
+\q
 {% endhighlight %}
 
 Now connect as a superuser that can install extensions:
 {% highlight bash %}
-$ psql
+$ sudo -u postgres psql
 postgres=# \c temba
 You are now connected to database "temba" as user "psql".
 temba=# create extension postgis;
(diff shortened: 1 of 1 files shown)


_docs/development.md
```diff
@@ -25,23 +25,26 @@ You'll need the following to get started:
 ## Create temba user for PostgreSQL
 
 {% highlight bash %}
-$ createuser temba --pwprompt -d
+$ sudo apt-get install postgresql-client postgresql-server-dev-9.3 postgis*
```
nicpottier: So I don't think we want to prescribe one distro or way of installing things, I think this stuff should be covered under the prerequisites.

_docs/development.md
```diff
@@ -25,23 +25,26 @@ You'll need the following to get started:
...
+$ sudo -u postgres createuser temba --pwprompt -d
 Enter password for new role:
 Enter it again:
+$ sudo adduser temba
```
nicpottier: You're creating a system user for temba here to make authentication easier for postgres, but I don't think that is actually required.

_docs/development.md
```diff
@@ -25,23 +25,26 @@ You'll need the following to get started:
...
 {% endhighlight %}
 
 ## Create temba database, add PostGIS
 
 Create the database as temba user:
 {% highlight bash %}
-$ psql --user=temba postgres
+$ sudo -u temba psql --user=temba postgres
 postgres=> create database temba;
```
nicpottier: I don't think sudo'ing as a system user is required here.

_docs/development.md
```diff
@@ -58,8 +61,15 @@ Now clone the RapidPro repository and link up the development settings:
...
-$ ln -s temba/settings.py.dev temba/settings.py
+$ cd rapidpro/temba
+$ ln -s settings.py.dev settings.py
+{% endhighlight %}
+
+At this point, you should edit the rapidpro/settings.py to adjust the password for the postgresql connection.
+
+##Install Node 
+{% highlight bash %}
+$ sudo apt-get install node node-less coffee-script
```
nicpottier: Again, this belongs in the prerequisities I think. Maybe we can just add something about "you should be able to type lessc and have it work".

We don't want to get into maintaining instructions on how to install the latest version of all our dependencies, rather link off to those projects.

_docs/development.md
```diff
@@ -69,6 +79,7 @@ pinned dependencies for RapidPro can be found in ```pip-freeze.txt```. You can
 build the needed environment as follows (from the root rapidpro directory):
 
 {% highlight bash %}
+$ sudo apt-get install python-virtualenv postgresql-server-dev-9.3 python-dev ncurses-dev
 $ virtualenv env
```
nicpottier: This is installing into the system python libraries, which we probably don't want to recommend. (the one exception being virtualenv, the others are part of pip-freeze and shouldn't be required)

nicpottier:
Apart from those few comments looks good.

_docs/development.md
litlfred: the build was failing when it couldn’t find the libraries.   is there perhaps an issue on the pip-freeze side of things?
Cheers,
-carl

_docs/development.md
litlfred: it failed for me w/o it.  maybe there was some other issue?
Cheers,
-carl

_docs/development.md
litlfred: right, but for those who have not worked with postgres before, they would need at least a link to the relevant docs
Cheers,
-carl

_docs/development.md
nicpottier: Maybe, but that would be an issue with how you have Postgres installed. The problem is that a lot of these little things are dependent on your OS, how you have the DB installed and configured etc, that's why I'm hesitant to put in some of these changes because they are specific to your environment. (ie, these work as written on our environments)

litlfred: I guess there are two options one is to list dependencies and let people try and figure it out.  The second is to say here are the install instruction for such-and-such distribution (e.g. Ubuntu 14.04 LTS).  For people that don’t know their way around unix, this later is a good option - otherwise you have a barrier for people getting RapidPro up and running.

If you say that you are only going to provide instal for LTS releases, you won’t have that much of a maintenance burden.
 
Cheers,
-carl

nicpottier: Ya, I think we are leaning towards the 2nd option at the minute, as it covers people installing on OS X and everything else as well. Otherwise it is too much of a moving target.

You are going to have to be familiar with Linux to run RapidPro, no real way around that, this isn't Word. :)

Issue #146, opened 2015-03-23
Webhook authentication in headers


ewheeler:
Many webservices (... like RapidPro's API) use headers for sending authentication/authorization credentials. 

Allowing the configuration of authentication/authorization headers will enable faster and easier integration with external systems.

ewheeler:
Org webhooks can now include user-defined headers (webhooks for org events configured on the Account Details page).

I think it would be valuable to add the ability to configure headers for Flow webhooks as well. This might sound crazy, but a valuable use-case I found using two flows:
1. a daily flow puts its respondents into several groups based on their responses
2. a flow later in the day sends an email summary of the day's group membership counts to relevant stakeholders

This could be implemented if flow number 2 could use a webhook (with auth headers) to lookup the group membership by hitting https://api.rapidpro.io/api/v1/groups.json?name=GroupA , https://api.rapidpro.io/api/v1/groups.json?name=GroupB , etc
It would also be amusingly meta to have rapidpro hitting its own API (aka rapidouroboros)

@nicpottier you mentioned [here](https://github.com/rapidpro/rapidpro/pull/149#issuecomment-85566727) some potential changes to Flow webhooks. any further ideas on that front now that rule-chaining is possible?

nicpottier:
![yo dawg](https://i.imgflip.com/oa584.jpg)

So I don't think the group thing would actually work, since you'd quickly get into paging. There's also a part of me that is absolutely terrified at what people would do with it. (restart the contact down a flow, creating a loop?)

All that doesn't really mean that I'm not ok with headers, just maybe not ok with API calls. We are about to ship some big UI changes this weekend on the flow editor and I believe @ericnewcomer has thought a bit about where webhooks would belong there. So once that is out and stabilized this would be more easily possible.

ericnewcomer:
First, I'd really like to +1 the yodawg.

I think what @ewheeler is describing would work. He's doing a separate call for each group which shows the membership count, so paging shouldn't come into play.

I think we do want to have header configuration in the flow webhooks. Yes, those dialogs are going to look a little bit different in the coming week. However, before we go too far down that path, we'll want to remove the webhook action since we really only want one way to do this (instead only using a webhook ruleset). That has some migratory implications, but should be manageable.

I share the same fear of rapidpro calling rapidpro creating the opportunity of an infinite loop, but that exists today, there just has to be an external server involved to accomplish it. Preventing that would require quite a bit more thought either way.

[...]

PR #149, opened 2015-03-24
add ability to configure headers for webhooks

ewheeler:
*Work in progress*
opening PR for discussing approaches to issue #146


ewheeler:
Adding the ability to customize a single header field name and value for Org webhooks is pretty straightforward. For Flow webhooks, this will be a bit more involved, will add a lot of noise to the modal dialog, and starts to feel less DRY.

A single Flow might hit the same webhook endpoint several times and an org might use the same webhook endpoint in several related flows. If the endpoint requires auth headers, thats a lot of copy-and-pasting of credentials and lots of places to make mistakes.

One idea: allow configuration/management of webhook endpoints independently of Flows. Then the modal could show the existing dialog and/or allow users to select from existing endpoints. The Flow and Org webhook config forms (and models?) could be the same.
Thoughts?

nicpottier:
Ya, I was going to mention that this Org webhook where we push events isn't used by very many people, and when it is it is used it is almost always just for messages where people are using us a fancy relayer. (I think there is strong argument to remove the other webhook event types here as a matter of fact)

So I don't think the Org webhook and the Flow webhooks are going to overlap much.

I'm also not sure that we see the same exact webhook called very often in flows as people generally are doing different things in different flows, though as you note the headers used for authentication probably would likely overlap.

I do like the idea of having a place to manage webhooks, wonder what the tradeoff there is between having to hit another page to go create your webhook before you can use it vs being able to configure it right in the flow. I guess the question there is more how many Orgs have more than one webhook action that is exactly the same. If most orgs use different webhooks in different flows, then having a different webhook configuration page will be annoying for them as opposed to configuring them inline, if most orgs use the same webhook in many flows, then it would be convenient.

From the queries I can run across both TextIt and RapidPro it sure looks like by far the most common case is for an Org to only use a webhook once, so that would put my bias towards doing it inline as opposed to having a separate place where webhooks live. I'm not mega passionate about it though.

@ericnewcomer Any thoughts?

ericnewcomer:
My primary concern is that making webhooks external creates another system
level dependency on flow definitions. So we'd need to decide how to handle
webhook deletion on flows that depend on them. Also not sure how we want to
handle export of those webhooks. It might not be immediately obvious to the
user that their export might include details about their off flow webhook
such as an api token.

Those issues aside, we could let the user configure or reuse a webhook in
the flow editor itself without too much complexity.

nicpottier:
Another quick consideration is that we are playing with the idea to change flow structure a bit in a way that might unify the webhooks in a Flow. (ie, there wouldn't be a webhook on rules as well as a webhook on actions, but only one on rules and the ability to chain rules) This comes with a bunch of nice side effects like being able to route based on whether your webhook returned a 200 or had an error etc...

This might be an argument towards just stuffing the header configuration in flows as at least there won't be repeats (in RapidPro code) of that across actions and rules.

ericnewcomer:
Also, if we add header support (which I feel we should), it should allow an arbitrary number of headers and probably doesn't need to be modeled on the db.

nicpottier:
Ya, concur that maybe this should be a JSON blob of the URL, headers and method.

ewheeler:
Thanks for the feedback!

@ericnewcomer agree with your concerns about adding dependencies on flow definitions and that having webhook endpoints outside of the flow editor doesnt make sense

@nicpottier chaining rules would be great. ability to route based on webhook response unlocks a lot

We've got a need for adding headers to Org webhooks coming up in the next couple weeks (would be nice to have Flow webhooks as well, but we can filter/ignore on the other side if we have the full stream). Given this discussion it sounds like punting on the Flow webhooks makes sense for now if there are other changes in the works -- and you don't foresee much overlap between the Flow and Org webhook implementations.
On the Org webhook side, JSON blob probably makes sense too (instead of separate URL and header fields). Any ideas for a UI for the blob? Maybe a URL field and method dropdown (a la flows' webhook modal). Then have two side-by-side fields for headers, with the little plus icon (the one used for adding actions) for adding another pair of fields?

nicpottier:
Agree that just doing the org changes for now makes sense, and moving it over to a JSON dict also makes sense. (you'll need a data migration to roll forward old data in the process)

I'm not sure METHOD really makes sense on the Org webhooks given the amount of data they are shipping, I think these will always be POSTs. For the headers we tend to use the pattern of just adding / removing field rows based on whether there are any empty. IE, we show one row of key/value as empty, and once the user enters something in either, we dynamically add another row. If they make both the key and value input boxes empty, then we remove them. 

Also ok with a plus button if it is pretty though. :)

ewheeler:
Added commit that includes:
- migration of org model to use an hstore dict field for `webhook`
- UI changes to allow for users to add an arbitrary number of headers

Todo/Questions (advice pls? @nicpottier @ericnewcomer):
- not sure how the haml-to-html voodoo works, so the dynamic addition of additional `<input>`s on the webhook form is dodgy and not internationalized: https://github.com/ewheeler/rapidpro/commit/39ecdc8255557bc697d6bcad1d3d5e20703cc6bf#diff-49bafca337e0fd7c2e9aa39b7685288bR154

- tests.. when i try to run `manage.py test` i get an error about hstore not being installed when migrations are run on the new test db.

ewheeler:
@nicpottier @ericnewcomer thanks for the feedback. pls review changes

temba/orgs/migrations/0002_auto_20150324_1044.py
```diff
@@ -0,0 +1,26 @@
...
+            field=models.CharField(help_text='Optional header field name to include with Webhook requests', max_length=255, null=True, verbose_name='Webhook Header Field Name', blank=True),
+            preserve_default=True,
+        ),
+        migrations.AddField(
+            model_name='org',
+            name='webhook_header_value',
+            field=models.CharField(help_text='Optional header value to include with Webhook requests', max_length=255, null=True, verbose_name='Webhook Header Value', blank=True),
+            preserve_default=True,
+        ),
+    ]
```
nicpottier: Since we've never pushed these migrations, better to remove the world where header_field_name and header_field_value every existed.

temba/orgs/models.py
```diff
@@ -563,6 +566,20 @@ def get_schemes(self, role):
         setattr(self, cache_attr, schemes)
         return schemes
 
+    def get_webhook_url(self):
+        """
+        Returns a string with webhook url.
+        """
+        return json.loads(self.webhook).get('url', '') if self.webhook else ''
```
nicpottier: I'd prefer None to empty string here.

temba/orgs/views.py
```diff
@@ -995,13 +995,13 @@ class CreateLogin(SmartUpdateView):
...
-            
+
             org = self.get_object()
             if not org:
                 messages.info(request, _("Your invitation link is invalid. Please contact your organization administrator."))
                 return HttpResponseRedirect(reverse('public.public_index'))
             return None
-        
+
         def pre_save(self, obj):
```
nicpottier: What editor are you using? looks like it is replacing these with tabs or something?

temba/orgs/views.py
```diff
@@ -1232,22 +1232,38 @@ class Webhook(InferOrgMixin, OrgPermsMixin, SmartUpdateView):
...
+            # grab all submitted headers
+            # discard any key with a falsy value or whitespace value
+            headers = {k: v for k, v in self.form.data.items() if (k.startswith('header_') and bool(v.strip()))}
+
+            # ensure each header _value_ has a corresponding key
+            # (technically, its ok to have a valueless header key)
+            for n in xrange((len(headers) / 2)):
+                pair = {k: v for k, v in headers.items() if k.startswith('header_' + str(n + 1))}
+                if len({k for k in headers if 'key' in k}) == 0:
+                    raise ValidationError(_('Headers must include a key'))
```
nicpottier: This validation probably belongs in the Form, not the View.

temba/orgs/views.py
```diff
@@ -1260,6 +1276,22 @@ def pre_save(self, obj):
...
+                webhook_data.update({'url': data['webhook']})
+                webhook_data.update({'method': 'POST'})
+
+            if data['headers']:
+                pairs = []
+                for n in xrange(len({k for k in data['headers'] if 'key' in k})):
+                    pair = {k: v for k, v in data['headers'].items() if k.startswith('header_' + str(n + 1))}
+                    pairs.append(pair.items())
+                as_dict = dict([(p[0][1], p[1][1]) for p in pairs])
+                webhook_data.update({'headers': as_dict})
```
nicpottier: Man this just looks super awkward in the few places I've seen it..  

If I understand things correctly we are just trying to go from header_n_key and header_n_value to a dictionary?  What about something like the below in the form to turn this into the dict.. 

Also any reason the Form isn't just taking care of the dict() transformation so you don't need to do this mapping twice?

```python
idx = 0
headers = dict()
key = 'header_%d_key' % idx
value = 'header_%d_value' % idx

while key in data:
  if data.get(value, ''):
    headers[data[key]] = data[value]

  idx += 1
  key = 'header_%d_key' % idx
  value = 'header_%d_value' % idx
```

temba/orgs/views.py
ewheeler: VIM is just deleting extraneous whitespace (removing eight spaces on an otherwise empty line in this case)

temba/orgs/views.py
ewheeler: Yup. agree

temba/orgs/views.py
ewheeler: good call. this is much cleaner

ewheeler:
@nicpottier please take one last look!

temba/orgs/migrations/0005_auto_20150416_0729.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('orgs', '0004_auto_20150416_0728'),
+    ]
+
+    operations = [
+        migrations.RunPython(cast_json),
+    ]
```
nicpottier: No real need to change this, mostly just an aside, but the new Django migrations are neat in that you could put this piece in the previous migration, which would then have everything run in a single transaction, so everything would switch at once.

temba/orgs/views.py
nicpottier: K, ya, might be nice to see if this can be switched off for the future just to keep these diffs simpler.

nicpottier:
Looks good, but but, where be the tests?

temba/orgs/migrations/0005_auto_20150416_0729.py
ewheeler: cool -- good to know.

ewheeler:
@nicpottier they're coming! haven't been able to run them yet.. as i mentioned above, when i run `python manage.py test`, i got an error while the migrations are running on the newly created test db. 

the error complains that hstore is not installed. 
i don't see anywhere that the hstore extension is added in a migration, so wondering what magic is necessary to get the test suite to execute

nicpottier:
Oh, if I recall you need to modify your postgres template db to have the right extensions. (which will end up being hstore, postgis and postgis_topology)

This has more details:
  https://docs.djangoproject.com/en/1.8/ref/contrib/gis/testing/

But essentially the gist is that when running the tests Postgres will use the 'template' database as a base, so you need to modify it to have the right extensions. Right now I'm pretty sure we default to using the 'template' database, not the 'post_gis' one, so you just need to do a psql, use that db and load those extensions.

ewheeler:
@nicpottier ok I'll check that out.
Any reason for not including in the migrations?

nicpottier:
Good question.. I'm actually not sure, it seems the "Django" way is to use the postgres template, but perhaps that is a vestige of the pre-1.7 days when they didn't have migrations built in.

Seems like that could be worthy of some investigation, making it easier to run tests seems like an obvious win.

ewheeler:
@nicpottier there may be more than one django way in this case: docs include an example of this at the bottom of this page https://docs.djangoproject.com/en/1.8/ref/migration-operations/

nicpottier:
Ya.. I like it, let's do it.

We need to put it in fairly early is all.. maybe add it to 0001_initial in temba/orgs?

nicpottier:
Actually looks like that one depends on locations, so maybe location is the right place to put the postgis stuff, and orgs the right place to put the hstore extension.

ewheeler:
@nicpottier sounds good.
what do you think about putting the `LoadExtension` class in `temba/utils/migrations.py` and then import that in `temba/locations/migrations/0001_initial.py` for postgis stuff and also import in the Org migrations that will load hstore?

nicpottier:
I don't feel super strongly either way, but my bias would probably be to do it inline instead of creating a new LoadExtension, just because it won't be used that often and it isn't much code.

nicpottier:
So to test this you need at least two tests:

1) To test the django form side of things, so that with various form submissions the org webhook dict is being updated correctly. Use this as inspiration:
   https://github.com/rapidpro/rapidpro/blob/master/temba/orgs/tests.py#L141

2) To test that given a set webhook configuration, calls are being made with the right headers, this one is a bit trickier as you need to use mock, but these will be similar:
   https://github.com/rapidpro/rapidpro/blob/master/temba/api/tests.py#L3521

nicpottier:
So APIAction isn't a model, it's a Class. It represents the action in a flow of calling out to a webhook.

nicpottier:
So ya, agreed having a more complete log of what was sent is ideal. I think there are probably reasons to maintain WebhookResult.url() just as it is, but I think adding something like 'request' which is a Text field that contains the raw ascii request makes a lot of sense.

Looks like you can get that from your prepared request with a few tweaks:
  http://stackoverflow.com/questions/20658572/python-requests-print-entire-http-request-raw/23816211#23816211

ewheeler:
Thanks, @nicpottier 

Cool. Leaving `WebHookResult.url` definitely makes things easier, bc thats used for the Flow webhook requests. Added a field for `WebHookResult.request`

Are yall planning a merge soon? http://github.com/nyaruka/rapidpro is quite a bit ahead of http://github.com/rapidpro/rapidpro and a few of the test errors I'm getting look to be resolved in http://github.com/nyaruka/rapidpro

ewheeler:
Thx for the merge @nicpottier 

I've got the green light from travis now, so plz review

temba/api/tests.py
```diff
@@ -3985,16 +3986,16 @@ def test_call_deliveries(self):
...
-            data = kwargs['data']
-            self.assertEquals('+250788123123', data['phone'])
-            self.assertEquals(call.pk, data['call'])
-            self.assertEquals(0, data['duration'])
-            self.assertEquals(call.call_type, data['event'])
+            prepared_request = args[0]
+            self.assertEquals(self.channel.org.get_webhook_url(), prepared_request.url)
+
+            data = parse_qs(prepared_request.body)
+            self.assertEquals('+250788123123', data['phone'][0])
```
nicpottier: Why did these all change? (the addition of [0]?)

temba/api/models.py
```diff
@@ -438,8 +454,11 @@ def record_result(cls, event, result):
             message = message[:255]
 
         api_user = get_api_user()
+
         WebHookResult.objects.create(event=event,
                                      url=result['url'],
+                                     # Flow webhooks won't have 'reqeust'
```
nicpottier: typo here. :)

temba/api/tests.py
nicpottier: Oh I see, this is because of using Session.send.. gotit.

nicpottier:
This looks good, want me to merge it?

ewheeler:
@nicpottier cool. go for it!

PR #155, opened 2015-03-31
"before" should be "before" and not "now or before"

xkmato:
I am wondering why api argument "before" returns a queryset "__lte=before" instead of "__lt=before" and "after" returns a queryset "__gte=after" instead of "__gt=after". 
I found this very confusing. I am wondering whether there is a specific reason behind it that I should be considering. thanks

### DIFF
--- a/temba/api/views.py
+++ b/temba/api/views.py
@@ -522,7 +522,7 @@ class BroadcastsEndpoint(generics.ListAPIView):
         if before:
             try:
                 before = json_date_to_datetime(before)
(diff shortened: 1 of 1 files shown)


nicpottier:
Although I appreciate you might disagree, we can't change API behavior at this point, at least not without changing all the parameter names. Will keep a note of this if we change our API version though.

xkmato:
I agree. I think that makes a lot of sense. I think what would be more
useful now is to make a note in the docs and in the rapidpro-python module
docs

PR #165, opened 2015-04-21
Split api send tests

jerith:
Almost all of the `test_send` methods in the `api.test` module currently fail due to duplicate send prevention. This branch splits each of those tests into two methods (one for success, one for failure) to avoid sending the same message multiple times and to keep the message state for the two cases separate.


nicpottier:
Hrmm, these don't actually fail for us. Wonder what is different there. 

How are you running them?

jerith:
I run `python manage.py test` after following http://rapidpro.github.io/rapidpro/docs/development/ and adding the postgis extension to the default database template. I see now that the `settings.py` I'm using overrides `REDIS_DB`, but `clear_cache()` hardcodes `db=10`.

This split is probably worthwhile in any case, though, because the success and error cases are different behaviours that shouldn't really share message state.

nicpottier:
Ok sure.. maybe worth putting the various channel setup bits in setup() for these now that we are repeating those bits on more time?

jerith:
I've factored out the channel setup (and a couple of other things common to all channel type tests) into a new base class and cleaned up a few inconsistencies in the process.

temba/api/tests.py
```diff
@@ -1803,9 +1813,7 @@ def test_send(self):
         # our outgoing sms
         sms = bcast.get_messages()[0]
 
-        try:
-            settings.SEND_MESSAGES = True
-
+        with self.allow_message_sending():
```
nicpottier: Nice, much cleaner.

temba/api/tests.py
```diff
@@ -3166,8 +3183,15 @@ def test_send(self):
...
                 self.assertTrue(msg.sent_on)
 
-                self.clear_cache()
+    def test_send_error(self):
+        # This tests a send error.
+        joe = self.create_contact("Joe", number="+250788383383", twitter="joe1981")
+        bcast = joe.send("Test message", self.admin, trigger_send=False)
+
+        # our outgoing message
+        msg = bcast.get_messages()[0]
```
nicpottier: Seems like these three lines are a good candidate to pull into the base class?

nicpottier:
This looks good, for extra credit would be nice to pull out those repeated lines as well, but I'm fine as is, that isn't your mess. :)

Can you send us a signed contributor agreement?

temba/api/tests.py
jerith: I thought about it, but there are lots of differences between the various `self.create_contact()` and `joe.send()` calls. Factoring them out into a method on each test class doesn't really buy very much. For the tests already had them in `.setUp()` I moved them to `.setup_channel()`, but I'm not sure they really belong there. (If they do, the method should probably be renamed.)

nicpottier: Fair enough. Admittedly these are some serious utility tests that were generally written in a rush as we were integrating a new partner, so ya, they could use some more love to normalize it all.

But you are definitely leaving them better than you found them. :+1:

nicpottier:
PS. And if you merge in rapidpro/master should be able to get Travis giving us a build status.

nicpottier:
Closing as it doesn't seem CAA is being signed.

Issue #163, opened 2015-04-21
Test environment set up


rudigiesler:
I'm having a bit of trouble setting up a test environment and getting all the tests to pass. I haven't been able to find instructions on getting a test environment set up in the documentation.

The steps taken are as follows:

1. Follow the documentation to get a dev environment set up (Able to run the server)
2. Add the extensions `postgis`, `postgis_topology`, and `hstore` to the `template1` database.
3. Run `python manage.py test`

The testing fails with 9 errors and 14 failures.

Travis-CI config: https://github.com/praekelt/rapidpro/blob/feature/issue-3-add-travis-ci/.travis.yml
Output of test failure: https://travis-ci.org/praekelt/rapidpro/builds/59359519

ewheeler:
hey @rudigiesler 
i noticed this too. theres a related discussion here: https://github.com/rapidpro/rapidpro/pull/149#issuecomment-93982541

and this commit will add these extensions via migrations to make it easier to run tests: https://github.com/ewheeler/rapidpro/commit/275e4eeb8e26898fa5877f9dbec2908e7497f354

nicpottier:
@ewheeler It looks like he knows the extension piece as rudi mentioned updating the postgres template.

@rudigiesler I wonder if your branch is just out of date. I'm trying your Travis file on our RapidPro fork to see if I get similar errors.

rudigiesler:
@nicpottier Merging in the latest changes from rapidpro/master doesn't seem to have any effect. The same tests are still failing locally for me.

nicpottier:
Hey @rudigiesler ..

I played with this a bit and got a build passing here:
  https://travis-ci.org/nyaruka/rapidpro/builds/59406120

A few things changed:
 * I'm not running the temba/perf_tests, which do have an error in them, but really we don't want running on every build because they are slow
 * I put in a step to symlink the dev settings.
 * To get a cleaner output from coverage, we're being explicit about the tests run. You can check the .travis.yml in nyaruka/rapidpro (used yours as a base) to see the packages we're running.

I'll get the perf test passing in a sec and push that to rapidpro as well.

rudigiesler:
Issue resolved in https://github.com/rapidpro/rapidpro/pull/166

PR #166, opened 2015-04-21
Adds Travis Integration File, Fix perf_tests

nicpottier:
Should fix issues in #163 encountered by @rudigiesler

Rudi if you want to review this that'd be swell.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ before_script:
   - psql -d 'template1' -U postgres -c "CREATE EXTENSION postgis_topology;"
   - psql -d 'template1' -U postgres -c "CREATE EXTENSION hstore;"
   - psql -U temba postgres -c "CREATE DATABASE temba;"
-  - ln -s temba/settings.py.dev temba/settings.py
+  - ln -s $TRAVIS_BUILD_DIR/temba/settings.py.dev $TRAVIS_BUILD_DIR/temba/settings.py
 script:
   - python manage.py test --with-coverage
 after_success:
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,6 @@ before_script:
   - psql -U temba postgres -c "CREATE DATABASE temba;"
   - ln -s $TRAVIS_BUILD_DIR/temba/settings.py.dev $TRAVIS_BUILD_DIR/temba/settings.py
 script:
-  - python manage.py test --with-coverage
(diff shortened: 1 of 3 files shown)


nicpottier:
PS. Travis log here:
  https://travis-ci.org/nyaruka/rapidpro

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+before_script:
+  - psql -U postgres -c "CREATE USER temba WITH PASSWORD 'temba';"
+  - psql -U postgres -c "ALTER ROLE temba WITH SUPERUSER;"
+  - psql -d 'template1' -U postgres -c "CREATE EXTENSION postgis;"
+  - psql -d 'template1' -U postgres -c "CREATE EXTENSION postgis_topology;"
+  - psql -d 'template1' -U postgres -c "CREATE EXTENSION hstore;"
+  - psql -U temba postgres -c "CREATE DATABASE temba;"
+  - ln -s $TRAVIS_BUILD_DIR/temba/settings.py.dev $TRAVIS_BUILD_DIR/temba/settings.py
+script:
+  - coverage run --source="." manage.py test temba/api temba/assets temba/campaigns temba/channels temba/contacts temba/flows temba/ivr temba/locations temba/msgs temba/orgs temba/public temba/reports temba/schedules temba/triggers temba/utils temba/values --verbosity=2 --noinput && coverage report -m --include="temba/api/*,temba/assets/*,temba/campaigns/*,temba/channels/*,temba/contacts/*,temba/flows/*,temba/ivr/*,temba/locations/*,temba/msgs/*,temba/orgs/*,temba/public/*,temba/reports/*,temba/schedules/*,temba/triggers/*,temba/utils/*,temba/values/*" --omit="*/migrations/*,*/tests" -i
```
rudigiesler: I see we're still not running perf_tests on travis here?

rudigiesler: There's an `after_success` config for travis that we can use here instead of the `&&`, which would allow us to split this line up. http://docs.travis-ci.com/user/build-configuration/#Build-Lifecycle

.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+after_success:
+  - coveralls
```
rudigiesler: New line needed at the end of the file.

README.md
```diff
@@ -1,4 +1,4 @@
-   
+
```
rudigiesler: Was this new line accidentally added?

rudigiesler:
Is it possible to get this repo hooked up to travis as well?

jerith:
Maybe the performance tests should run in a separate build job? Alternatively, perhaps they could run in the normal build but only if all other tests pass?

nicpottier:
Ya, doing the coverage step in ```after_success``` is definitely the right call.

The perf tests don't really do anything useful at the minute, so I'm not worried about putting them in, they haven't been touched in over a year and I'm not sure even relevant, so until they are I don't think having them in the Travis build makes sense.

We do however have some new automated browser tests for the Flow builder, and that we'll get hooked in next week.

rudigiesler:
The problem with leaving tests out of travis is if they never get run, then they will get out of date and fail as the code changes. Maybe they need to be deprecated and removed in the near future? Since they seem abandoned. Maybe an issue to decide on what to do with these tests going forward?

Otherwise I'm :+1: with the latest commits.

nicpottier:
@rudigiesler Oh ya, no totally agree on that point, as demonstrated with this whole fiasco in the first place by those being broken.

Opened #167 to track the perf tests.

nicpottier:
@rudigiesler Can we get a contributor agreement from you as well? (since this was really your work?)

PR #182, opened 2015-07-15
add docs about advanced settings in relayer

ewheeler:
document how to use the [android sms channel app](https://github.com/rapidpro/android-channel) with a development server


nicpottier:
Nice. Been chatting with Matt about this a bit. Might want to add that you can set the port to something else by specifying it in the IP address, ie: 192.168.1.15:80 

Thanks!

ewheeler:
@nicpottier -- good call on specifying the port -- i've added a sentence on specifying another port.

I've also included a paragraph on the message packs so we've got that in the docs too. Not sure if it really belongs in `_docs/development.md` but I'm not sure that it fits any better or worse in any of the other files in `_docs` either...

nicpottier:
Woot.

PR #183, opened 2015-07-16
add contact language substitution

ewheeler:
Adds functionality so that ```@contact.language``` is available for autocomplete in the Flow editor and the accompanying substitution/evaluation happens

Due to the ordering in [```temba/contacts/models.py```'s ```build_message_context```](https://github.com/rapidpro/rapidpro/blob/master/temba/contacts/models.py#L912), if orgs have a pre-existing ```ContactField``` called "language", then that value will be displayed instead. This is not ideal if any such pre-existing ```ContactField```s do in fact exist in production.
 
Options for avoiding this:

1. Re-order ```build_message_context``` such that the ```RESERVED_CONTACT_FIELDS``` are added to the dict after any ```ContactField```s and thus will replace any pre-existing value for 'language' from a ```ContactField```

2. Data migration to copy any values from pre-existing ```ContactField```s named "language" to the ```Contact``` model's ```language``` (if they are an iso3 that the org is configured to use...) and then delete the pre-existing ```ContactField```


nicpottier:
Ooh, that's a good catch.. I wonder if the expansion should be to the language name instead of the iso code though? I guess it won't be localized so maybe that isn't that useful.

What is your use case for this out of curiosity? Maybe that would inform the decision..

ewheeler:
@nicpottier Use case is to be able to include the language code in webhook requests. We had a webservice that was returning data as well as the next message to be sent to the Contact. Otherwise you've got to branch on contact's language and 'hardcode' the language param in the webhook url

Also for clarification, am I understanding correctly that the intent of the assertion [here](https://github.com/rapidpro/rapidpro/blob/master/temba/msgs/tests.py#L126) is to confirm that its a single object. Both lists here have the same contents, but not the same order

nicpottier:
Ok, I'm ok with language code then.. I think the biggest argument there is that users will probably have their own localized name for the language rather than ours, so wouldn't be adding much value there.

Are you sure the content is the same? Seems like the test completions would be missing ```contact.language``` now. Ordering should be the same as these are lists.

ewheeler:
@nicpottier yeah i added ```content.language``` to the test completions. the content is the same in the travis output and my local output, but not the order

temba/msgs/tests.py
```diff
@@ -114,6 +114,7 @@ def test_send_message_auto_completion_processor(self):
                        dict(name='contact.tel_e164', display="Contact Phone - E164"),
                        dict(name='contact.groups', display="Contact Groups"),
                        dict(name='contact.uuid', display="Contact UUID"),
+                       dict(name='contact.language', display="Contact Language"),
```
nicpottier: This should be alphabetical, pretty sure that's why it is failing.

temba/msgs/views.py
```diff
@@ -43,6 +43,7 @@ def send_message_auto_complete_processor(request):
         completions.insert(4, dict(name='contact.tel_e164', display=unicode(_("Contact Phone - E164"))))
         completions.insert(5, dict(name='contact.groups', display=unicode(_("Contact Groups"))))
         completions.insert(6, dict(name='contact.uuid', display=unicode(_("Contact UUID"))))
+        completions.insert(6, dict(name='contact.language', display=unicode(_("Contact Language"))))
```
nicpottier: Ditto.

temba/flows/views.py
```diff
@@ -686,6 +686,7 @@ def render_to_response(self, context, **response_kwargs):
                 dict(name='contact.tel_e164', display=unicode(_('Contact Phone - E164'))),
                 dict(name='contact.uuid', display=unicode(_("Contact UUID"))),
                 dict(name='contact.groups', display=unicode(_('Contact Groups'))),
+                dict(name='contact.language', display=unicode(_('Contact Language'))),
```
nicpottier: For consistency might as well order these alphabetically as well.

temba/msgs/views.py
ewheeler: I think I'm missing something in terms of alphabetizing this, unless this test was already failing. You've got a strange and mysterious alphabet here if ``tel_e164``` or "Contact Phone" comes before either ```contact.groups``` or "Contact Groups"

Here, though, i botched the insertion and didn't increment the position

nicpottier: It's good to have SOME order rather than just have these be random, so let's just alphabetize them, at least you will know what the "right" behavior is then.

ewheeler:
@nicpottier completions are ordered alphabetically now and tests are passing

What are you thinking regarding the potential issue with previously-existing ```ContactField```s named "language"?

My feeling is that clobbering any existing ```@contact.language``` with the Contact's real language setting could only be a good thing. If ppl have defined their own ```ContactField``` then they probably aren't taking advantage of the i18n functionality as they intended to.

nicpottier:
Excellent.. ya, I'm actually ok shadowing user defined "language" fields.

Looking at our existing data, there are some RapidPro accounts with a "Language" contact field, so maybe why don't we write a migration that changes the label and key for "Language" fields to "User Language" so at least it is clear that those are different things now.

That won't change flows that reference those contact fields, as those are based on the key, but they all look like test accounts so should be ok. Looking at the accounts I don't think we'll break anything there.

ewheeler:
@nicpottier ok, i've added a data migration to rename existing ```ContactField```s named "Language"

nicpottier:
Nice.

Issue #68, opened 2014-12-16
Remove from All Groups Action


nicpottier:
It may be nice to have an action that removes a user from every single group without having to enumerate them one at a time. This is especially true in cases where groups are actually dynamic.

See: https://textit.uservoice.com/admin/tickets/404

ewheeler:
@nicpottier are you thinking that this would be a separate action? or some kind of button on the existing 'Remove from Groups' dialog and an additional execution mode for [```DeleteFromGroupAction```](https://github.com/rapidpro/rapidpro/blob/master/temba/flows/models.py#L3750) that selects all ```ContactGroup```s for the org?

nicpottier:
Could be part of the same action, though the UI seems slightly tricky since we use an omnibox there I think.

PR #188, opened 2015-07-28
Remove from all groups action

ewheeler:
Initial stab at resolving https://github.com/rapidpro/rapidpro/issues/68

Implements ```DeleteFromAllGroupsAction``` as a separate action, which seems a bit clunky. 

Probably better to combine with the existing [```DeleteFromGroupAction```](https://github.com/rapidpro/rapidpro/blob/master/temba/flows/models.py#L3750), but as @nicpottier [pointed out](https://github.com/rapidpro/rapidpro/issues/68#issuecomment-125201204) this will be tricky on the frontend bc of omnibox. 

This could look like a new option for the [omnibox widget](https://github.com/rapidpro/rapidpro/blob/master/static/coffee/flows/widgets.coffee#L451) that adds a suggestion of 'Remove from All Groups' to the top of the list when a user clicks into the omnibox or starts typing 'all' (maybe with a different or red icon)
del_groups_from_list

Alternatively, there could be a 'Remove from All Groups' checkbox next to the "Choose one or more groups below." label that would hide/disable the omnibox when checked.
del_groups_checkbox

I'm leaning towards the checkbox approach because 1) compared to the new dropdown item approach, i think this more clearly communicates to the user that this option does something different and 2) avoids mucking too much with the omnibox


rowanseymour:
Agree checkbox is definitely more intuitive.

nicpottier:
Eric is the UI nazi on the flow builder, so will wait for him to get back to comment. (Thursday)

I think both approaches are ok, though the checkbox also involves locking down the omnibox so people can't enter items there. (and clears existing ones) And those fonts need tweaking, but thanks for doing this Evan!

ewheeler:
cool i'll wait to hear from eric.

heh, yeah those are just mockups, so the font will be right once we've settled on a direction.

btw heres how the code in this PR looks (as a separate action in addition to DeleteFromGroup/del_group):
screen shot 2015-07-28 at 4 35 27 pm

ericnewcomer:
Definitely don't want to muck with the omnibox for this.

As Nic said, if when you check the checkbox it hides the omni (and vice-versa), then I think this is okay.  You'd probably want to have the Remove from all groups option come first (possibly on its own line) so the label for the omni can be hidden along with it when it is checked.

ewheeler:
cool. thanks @ericnewcomer 

I've removed the separate 'Remove from all groups' action and added a checkbox to the 'Remove from groups' action. I'm using ```ng-show``` and ```ng-hide``` to hide the omnibox and its description as well as showing a different 'Ok' button that calls a different controller (is this a sane approach for this? adapting the saveGroups controller to handle this additional case seemed unnecessary, but defer to you on this)

Here's what it looks like now:
screen shot 2015-08-04 at 5 18 53 pm
screen shot 2015-08-04 at 5 19 08 pm

On the second screenshot, you'll notice that the 'Ok' button is disabled. I can't figure out how to enable this when the checkbox is checked
I think I want a directive to do this (https://github.com/rapidpro/rapidpro/commit/adcf21ff68c7101f1a5bd5c0708b74c3b34c0bc5#diff-8f8966d5d5e7bf93c3c5f275da9ef3fdR397)? but not sure what it should do specifically or what magic the omnibox is summoning to enable the button for submission

ericnewcomer:
The problem is that the group selection is required for the form submission. Normally this would be done with a flag on ng-required set according to the view state. Since it is using the omnibox / select2 directive (which handles requiredness itself internally) I think there are two options:

1) You can also manage visibility with ng-if. I think doing that should remove it from the DOM and thus letting the form be valid when it isn't present. 

2) You can use the required flag for the omnibox widget and make it evaluate according to visibility. This isn't something that's been tested with the omnibox directive, so YMMV on this one.

On the L&F.. man, it looks pretty ugly. I guess you can blame me for bad suggestions there. Alignment definitely makes it feel goofy. Also, I feel the common operation is to remove from one group, so having it presented first really does feel odd now that I see it. Maybe keeping it exactly as it was originally but just adding a box underneath for the remove all case. No need for "or". I would also tweak the alignment so the checkbox lines up with the select widgets.

ewheeler:
Thanks for the tips @ericnewcomer 

I've got the action saving properly and working properly in test flows (meaning it works as advertised and removes contacts from all the groups)

Using ```ng-if``` did the trick for hiding the omnibox and allowing the form to submit with the box checked. Also updated the L&F which is a bit of an improvement:

screen shot 2015-08-05 at 4 11 43 pm

screen shot 2015-08-05 at 4 12 08 pm

In order to get the [actionName directive](https://github.com/rapidpro/rapidpro/blob/d058aa0cbaff5329cf6a310fded9a157b3e0f80b/static/coffee/flows/directives.coffee#L201) to display the action correctly, I added an array called [$rootScope.hiddenActions](https://github.com/rapidpro/rapidpro/blob/d058aa0cbaff5329cf6a310fded9a157b3e0f80b/static/coffee/flows/services.coffee#L397) and updated [getActionConfig](https://github.com/rapidpro/rapidpro/blob/d058aa0cbaff5329cf6a310fded9a157b3e0f80b/static/coffee/flows/services.coffee#L773) so the action name/icon are updated. This felt hacky, but I'm not sure what a preferred approach would be. From what I can tell, much of the flow editor code assumes a one-to-one relationship btw action types show to the user and associated controllers. I'm definitely bending that here, but haven't seen any terrible side effects yet. Not sure if this kind of bending will come up again or not, so hard to tell whether a larger refactor would be more prudent. Thoughts?

The big part thats left (and where I'm currently floundering) is getting the modal to pop up correctly when clicking on the 'Remove from All Groups' action in the flow editor. Currently it pops up a blank action modal: 
screen shot 2015-08-05 at 4 29 43 pm
I don't totally follow what the ```NodeEditorController``` is doing or where value of ```[[config.type]]``` is set for [templates/partials/node_editor.haml](https://github.com/rapidpro/rapidpro/blob/d058aa0cbaff5329cf6a310fded9a157b3e0f80b/templates/partials/node_editor.haml#L14)

Sorry for all the noob questions! theres a lot going on in the flow editor that I'm still getting up to speed with (esp. angular)

ericnewcomer:
I think we should leave the title as Remove From Groups. So, it should definitely still be the same action, just with a slightly different treatment for the body text if that flag is set. So, no need for mucking with the actionName directive and the hiddenActions stuff. That definitely doesn't smell right. With it being the same action, everything else should just work when editing, etc.

ewheeler:
@ericnewcomer cool. leaving the title is all good.

unless i'm mistaken-- without having this as a separate action, this feature could only really be an alternative to entering all of the org's currently existing groups into the omnibox. meaning that when the page reloads it will look and behave like the existing 'remove from groups' action with whatever groups were present in the org when the action was initially saved (and when you click the action to edit them you'll get the omnibox). if the org gets additional groups after you add the action--and the contact is put into those groups--then contacts won't be removed from the new groups.

i was imagining a 'remove from all groups' action that would always remove a contact from all of the groups that they are a member of each time they go through the action. if a user intends to remove a contact from all groups, then they shouldn't need to remember to edit this action each time there are new groups in their org that contacts might be in

unless its a distinct action, i don't see a way to implement a [get_groups](https://github.com/ewheeler/rapidpro/blob/d058aa0cbaff5329cf6a310fded9a157b3e0f80b/temba/flows/models.py#L3665) function for [DeleteFromGroupAction](https://github.com/ewheeler/rapidpro/blob/d058aa0cbaff5329cf6a310fded9a157b3e0f80b/temba/flows/models.py#L3751) that can behave in a way that accommodates both the existing functionality (fetch a specific set of groups) as well as the proposed new functionality (fetch all of the org's groups)

ericnewcomer:
I think you'd modify ```as_json``` and ```execute``` on DeleteFromGroupAction. It would only call ```get_groups``` if the flag isn't set to remove from all groups.

ewheeler:
hey @ericnewcomer could you take another look? this is working nicely now

PR #226, opened 2016-01-29
Update LICENSE

ewheeler:
its a new year :fireworks: :tada:


rowanseymour:
Was wondering when someone would tackle this bit of work

PR #229, opened 2016-02-05
USSD flow type + VUMI USSD Channel

nullcode:
The USSD flow type allows session based communication through a USSD channel. 

A USSD message is very similar to both text- and voice messages, but still has a different representation in terms of UI.
A USSD message is text based and can contain 182 characters. The most common use of USSD is by using menus with which users are able to easily navigate and answer questions by typing numbers or characters.
The difference between a USSD menu and a text message in terms of RapidPro logic is that the USSD message can contain rules in the message itself in the form of a USSD menu. A RapidPro flow consists of action and rule sets, so when we are speaking of a USSD menu it means that a USSD flow element is both an action and a rule item. Also, the USSD menu always waits for a response from the user (unless terminated). With this in mind chaining actions could result in a broken RapidPro flow. @nicpottier and @ericnewcomer suggested a solution where there would be no USSD action sets in a flow but only USSD rules that also contains the message and menu to be sent out.

ussd_ruleset_editor

The USSD message can contain: 
- text
- menu (that is also rendered to plain text with slight formatting - every menu item is in a separate row)
- arbitrary rules

The example above renders to the following USSD flow rule item:

ussd_flow_rule_item

Some important features to be acknowledged here:
- The rule item contains the question to be sent out with the menu
- The menu is rendered as a ruleset (categorised as "Sports" and "Politics")
- The arbitrary rule is also added to the ruleset (categorised as "Movies")
- "Other" rule is reserved for all other responses
- "Interrupted" rule is triggered when the USSD session has been terminated while the flow item was waiting for a response

Simulator: TBA

This feature is currently a Work In Progress, but is open for discussion. Any constructive feedback is welcome!


nicpottier:
I like the interrupted choice, that's nice and will be super useful.

On the UI side of things..

That dialog is a mouthful, what if we had two ruleset types "Wait for USSD Menu" and "Wait for USSD Response".

"Wait for USSD Menu" should just contain just the "Add Menu" options and would be the common case, automatically build the text for the menu cases etc.. 

The "Wait for USSD Response" would be the escape hatch for more advanced stuff, there the text displayed to the user would be completely in control of the user (no auto-menu creation) and only the arbitrary rules would be present in the dialog. ("If the message response.. ")

I think that provides the right balance of having a lightweight UI but still allowing anything to be built if need-be.

In the overview state, I don't think having special treatment for the menu options is a good idea, that's likely to cause confusion for users as what they see on handsets won't match it, the text of the USSD menu should be presented exactly as it will appear on the device.

Looks great though, looking forward to having this!

nullcode:
@nicpottier I support the idea of having two options and it actually simplifies the logic to be put in the directive too, so lets do that!

As for the flow item, we had several discussions about styling the menu or not and couldn't really pick one over the other so I decided to push this version and wait for feedback. I agree that it can be confusing, I'm down with changing it back to the normal representation. Then both the flow item and the simulator will show the message exactly how it will be presented on a phone.

nicpottier:
Awesome, sounds good and thanks!

nullcode:
added some tests to cover new functionality

here are some screens to show what it looks like (presented with the same example from above):

ussd_menu_response

ussd_response

and the flow items with connections between them:

ussd_flow_items

nullcode:
@nicpottier @ericnewcomer I have come to a point where I'd really appreciate your input.
The problem: Rulesets weren't meant to send messages (take actions), however the USSD ruleset is a special kind that needs the ability to send messages.

My first idea was to deal with this on backend level. The flow consists of FlowSteps where the type of the FlowStep can be either an ActionSet or RuleSet. So the general solution would be to treat the USSD RuleSet somewhat different and have a case where it would create outgoing messages (eg: Msg.create_outgoing as the ReplyAction has) where it reaches the flow step or perhaps create a UssdAction that will deal with it in a similar fashion. After I went for implementing this idea I had come across some problems with following the UI flow changes appropriately. I'd like to untangle this area, maybe you could point me to the right direction.

I was also thinking about creating a "shadow" actionset which would be on the UI flow, but hidden from the user (for eg: be on the y-1 coordinate above the ruleset, 1 unit wide and hidden). This would replicate the same action+ruleset combo and leave the system in a consistent state regardless the flow type.

What do you guys think? If possible, we could also have a chat session on Slack about this, I'm more or less available entire day, so you can propose a timeframe.

Thanks

ericnewcomer:
We discussed this a bit and think that having the RuleSet hold the message to be sent and have message sending abilities to be the best way to go. RuleSets have an arbitrary json config which you can use to store the messages for these USSD rulesets as needed. Just make sure you store your message data in a localized fashion like the ReplyAction does, etc and can be localized using the translation editor in the same way.

I don't think you'd need any "Action" like those used by ActionSets. Just a RuleSet that based on its type that can send a message (and uses its config to store the necessary prompts).

I'll be lurking in the slack channel if you want to come discuss it further.

nullcode:
I'm just about to merge the translation feature into this branch. With this I consider this feature final, all proposed functionality should work (except the USSD channel implementation itself) and testable with the flow simulator. Any feedback is welcome.

Also, can you guys check the tests on the master branch (I merged it in my branch and turns out it's failing) [https://travis-ci.org/rapidpro/rapidpro/builds/115939867]

Cheers

nicpottier:
Build should be fixed now. (had a DST bug we had fixed in our branch but not pushed upstream yet)

Will take a look at this over next few days. Note we can't merge anything that lowers coverage, so keep your eye on that as you continue work on this.

nullcode:
i pushed the final polishes for my work, so here's what we have:

PR includes:
- USSD Flow
- USSD Menu ruleset
- USSD Response ruleset
- USSD Simulator
- USSD Translator
- VUMI USSD Channel
- standard triggers for starting USSD flows

PR does not include:
- USSD Pull support (alias USSD triggers)
- USSD shortcodes
- USSD interrupted state

Note: Currently a USSD flow can be only run through a USSD channel. Although it makes sense to run it through other session based services like Messenger, Telegram (already tested working in dev), I've not enabled these for now. 

Note2: An other limitation is the VUMI USSD Channel itself, which apparently operates with South African USSD codes only, so in order to see the channel option, you need to set your organisation timezone to match.

templates/flows/flow_editor.haml
```diff
@@ -247,14 +251,24 @@
                 It's a good idea to start your call off by saying something first. You'll probably want to
                 introduce yourself and ask them a question. Just type the message for now, you'll be able to
                 add a voice recording of your message later.
+            -elif flow.flow_type == 'U'
+              -blocktrans
+                Welcome to a USSD flow. USSD is a session based communication type. We recommend starting your flow
+                with by sending a USSD message. This message will be sent to anybody right after they join the flow.
```
nicpottier: "your flow by"

templates/partials/ussd_directive.haml
```diff
@@ -0,0 +1,30 @@
...
+.split-header-inner
+  %p
+    Add text:
+
+.split-margin
+  %textarea{ng-model:"ruleset.config.ussd_text",name:"ussd",auto-complete:"",flow:"[[flowId]]",ng-change:"countCharacters()",class:"ussd-add-text"}
+
+.split-header-inner{ng-if:'USSD_MENU'}
+  %p
+    Add menu:
```
nicpottier: trans

temba/channels/models.py
```diff
@@ -61,6 +61,7 @@
 TWITTER = 'TT'
 VERBOICE = 'VB'
 VUMI = 'VM'
+VUMIUSSD = 'VMU'
```
nicpottier: ```VUMI_USSD``` matches the style we use better.

nicpottier:
This looks good at first glance, @ericnewcomer please take a look as well.

Just merged master, so might be a few migrations that need moving around.

I'd like to load a version of this on our master server so we can see how it feels, will do that by early next week.

nullcode:
@nicpottier thanks for the comments, I corrected them and solved all the merge conflicts

static/coffee/flows/services.coffee
```diff
@@ -459,7 +468,7 @@ app.factory 'Flow', ['$rootScope', '$window', '$http', '$timeout', '$interval',
         # { type: 'pause', verbose_name: 'Pause the flow', ivr:true, text:true},
       ]
```
ericnewcomer: Feels weird to have Split by message form be an option here. This is a feature for parsing delimited SMS forms which is also a "wait" ruleset. The only waits here should be USSD based.

ericnewcomer:
Do we still want to collapse like-named categories for USSD if they have distinct menu options?

screen shot 2016-04-21 at 12 44 42 pm

ericnewcomer:
Some minor styling nits.. 

- The borders on the left are consipring to create a double width to the left of the categories. 
- The padding for the top, bottom, and left for the menu should all be visually equal.

styling

ericnewcomer:
Missing text should not be possible when submitting the USSD dialog. In this example, I have an org with no languages set. It allows me to submit the USSD dialog with no primary message and a menu option with no text. It seems a primary menu message need not be required (but shouldn't show the translation treatment) and the menu option text should not allow the user to proceed without setting it. Either way, the missing-translation styling is improperly applied here.

screen shot 2016-04-21 at 12 51 46 pm

ericnewcomer:
Localization treatment should not have a white border.

screen shot 2016-04-21 at 1 01 56 pm

ericnewcomer:
- Translation dialog is unbalanced, way too much white space. Perhaps extend the inputs?
- The "Other" text is missing.

translation_dialog

The translation dialog in general has some whitespace issues, however, this moves in the wrong direction.

temba/flows/models.py
```diff
@@ -4412,6 +4441,68 @@ def execute(self, run, actionset_uuid, msg, offline_on=None):
...
+
+            return ussd_action
+        else:
+            return UssdAction()
+
+    def prepare_localised_msg(self):
+        # if there is a translation missing fill it with the base language
+        for language in self.languages:
+            if language not in self.msg:
+                self.msg[language] = self.msg[self.base_language]
```
ericnewcomer: In my tests self.base_language here looked to be using the primary language for the org ("eng"), not the base_language of the flow (which was still "base").

* Start with org with no languages
* Add English as the primary language to org
* Simulate flow

nullcode: nice catch, thanks

xkmato:
Hi guys,

I am planing to complete this PR. I will be adding two main things;

1. USSD Exit screen - Which is a USSD action that does not expect a response. This the the last screen sent after a USSD conversation. I will also allow rapidpro to move on to the next step. - For example if you would want to send a message after a USSD session. This action should help gracefully exit the USSD session
flow1 This is how the screen action will look like(Just like SMS action) I am not yet sure about the character limit for this
ussdclose 

2. The white space for USSD ruleset - I have decided to center the elements so then they look a slightly more organised. This also centers non USSD rulesets
movingtocenter
nonussdtrans

Let me know what you think so that I can proceed to add these. Cheers

temba/flows/models.py
ericnewcomer: Sorry, as I recall, the issue wasn't precedence. In my test I believe it was blowing up following those repro steps because the flow didn't have a value for the org language.

ericnewcomer:
No more @nullcode?

It looks like in your screenshot there is still no routing for interrupted sessions.

For closing a USSD session with an action, some thoughts..  The flow should be terminated in this case, ie, it should no longer be extendable. Also, as an action, you would want to disallow creating two of these actions in the same actionset. This is not dissimilar from the start a new flow action (which is a little bit goofy as well since the current flow run is terminated in that case).  This behavior probably deserves a fresh look, but I can be convinced extending the current pattern for starting flows is an acceptable next step here.

I'm okay with centering the translation dialogs.

xkmato:
Good point on the session interrupted option. I am not actually sure why that was taken out. But I think it should come back. I will need to get more familiar with the USSD APIs to add this.

About the exit screen(Closing USSD session), I think this is only useful if we are planning to extend the flow after the USSD session. Eg, When we want to send the summary via SMS from a USSD registration flow using the same flow. But from what I gather, the USSD flow is not expected to support SMS, am I right? I imagine a couple of cases where we would want it to though. But if we don't want to consider that, then I think the `Close USSD session` action is also redundant because `Wait for USSD Response` rule will already covers that. That also means that the only thing left to add would be to add the Session Interrupted Option

ericnewcomer:
I don't think it was removed, I think the work for that was never completed. Is puli no longer working on this?

nullcode:
We currently have no resources assigned to this, but we have a UNICEF LTA support in the works. We also have a roadmap what to work on next, like the interrupted state, USSD triggers (pull) and USSD short codes. As matter of fact I already have this figured out for VUMI, will need several days of implementation though.

nullcode:
AFAIK to merge this PR it would be good to have the Interrupted State feature included. In the first run it would be implemented for the VUMI USSD API + all the UI and logic behind interacting with the state. 

Is this in line with your preferences ? Is there anything else that I should look at in order to have this completed ?

ewheeler:
@nullcode thats my recollection

anything else you'd like to see before shipping this @ericnewcomer and/or @nicpottier ?

nicpottier:
Yep, along with the styling comments that's correct.

temba/flows/models.py
nullcode: gotcha, could reproduce and fixed in a4988799200fbe199660333e4aa89d4eccedde46

nullcode:
All styling issues should be fixed now along with @xkmato 's commit, also all the bugs found previously were terminated.

I added the Interrupt Rule feature for USSD. Please take a look, if you want to test it with the simulator just send the `__interrupt__` text in.

To summarise what this PR contains:
- USSD Flow
- USSD Menu ruleset
- USSD Response ruleset
- USSD Simulator
- USSD Translator
- VUMI USSD Channel
- USSD Interrupted state
- standard triggers for starting USSD flows

The following features (along with more USSD gateway API's) will be added next in a separate PR:
- USSD Pull support (alias USSD triggers)
- USSD shortcodes

ericnewcomer:
Had another pass through this and have a few comments.

* Looking at the flow definition this generates, it's not very DRY in that the rules are effectively duplicated in the config for the ruleset. One thought we had is maybe instead we add a "label" to rule which would remove the need to duplicate all that info. You could then just infer the option from the rule test.

* Create a Wait for Response ruleset instead of a menu. Save that. Then open it again. You'll see the interrupted rule listed erroneously.

* For the interrupted rule test type, let's use "interrupted_status"

* What is "ussd_text" and "ussd_message" in the ruleset definition? Seems they are the same thing? As such ussd_message seems like the one to keep as it is localized.

* For chars remaining, can we make that look closer to the counter for remaining characters on sms flows for consistency?

*  Interrupted should be styled just like the other rules (look at timeouts for sms flows as an example)

* Styling for simulator should either remain consistent with sms flows or break a bit more towards USSD. ie, the menu should probably not have a chat tail (tho the response should).

* Starting a USSD flow without a channel shouldn't be allowed, instead there should be text instructing them to add a ussd channel to start that flow. That should be the case if there are channels on the account but still no USSD channel.

nullcode:
Thanks for the comments @ericnewcomer 
I have fixed the bugs and I'm refactoring some of the parts based on your comments right now.
Some comments from my end:

> For chars remaining, can we make that look closer to the counter for remaining characters on sms flows for consistency?

It would break the UX if it was under the text area like SMS, because the menu items count in the message too. I'd suggest to either update SMS to this version or come up with an idea that suits both better. 

> Starting a USSD flow without a channel shouldn't be allowed, instead there should be text instructing them to add a ussd channel to start that flow. That should be the case if there are channels on the account but still no USSD channel.

I assume you meant to not start a flow without a channel added (like other flows). I added this now.

> I don't see a way to add a USSD channel

An other limitation is the VUMI USSD Channel itself, which apparently operates with South African USSD codes only, so in order to see the channel option, you need to set your organisation timezone to match.

ericnewcomer:
Understood on the character front. You are right, I agree it needs to stay below the menu declaration. Would still like to not have the numbers styled differently. Maybe change it to something like...

XX characters remaining

without any special styling there.

nullcode:
I dried it up and fixed the styling issues.

Let me note that this PR has been dragged out for too long. The biggest issue is when you guys upmerge I have enormous amount of merge conflicts every time. One of the last upmerges took me 2 days to merge into this (fix merge conflicts, introduce typos, fail to delete, fail to add some lines, hours of debugging, etc..) At this point it's not feasible anymore to work on a mammoth PR like this. The next PRs regarding USSD should be more minimal.

I hope and believe that all the issues have been fixed and all your preferences have been met regarding the features bundled into this PR and can be finally merged.

ericnewcomer:
Hey @nullcode looks good, merging this now.

As for merge conflicts, that's just the nature of a fast moving project with high standards. Trust that all of our PRs get the same scrutiny internally. Of course the larger the PR, the more problematic it is, so it certainly makes sense to bite things off in more manageable pieces.

sovello:
Hi!
Is there a documentation how to implement this?

ericnewcomer:
Hi @sovello, this really isn't the right place for that. Please join the RapidPro development google group.

https://groups.google.com/forum/#!forum/rapidpro-dev

PR #234, opened 2016-02-18
fix running karma tests with Chrome browser and debugging

nullcode:
Flow service "flow completion" fetch should not iterate through null variables.
It resulted in "Array.prototype.slice called on null or undefined" with Chrome launcher thus some tests failed.
This fix will ensure both PhantomJS and Chrome browsers to have same results.


ericnewcomer:
This looks great. We can merge this once we have the contributor agreement from you.

PR #238, opened 2016-02-26
Ward level implemetation

musamusa:
**Updated flow:**
 
- can create ward level flow (has ward)
-  will validate ward names
-  will fail if multiple ward names is returned from parse_location
-  can query district and ward levels without going through admin level hierarchy  (State->District->Ward)

**Updated Contact:**

- Can create ward


temba/contacts/models.py
```diff
@@ -362,8 +362,14 @@ def set_field(self, user, key, value, label=None):
             dec_value = self.org.parse_decimal(value)
             loc_value = None
 
-            if field.value_type == Value.TYPE_DISTRICT:
-                state_field = ContactField.get_state_field(self.org)
+            if field.value_type == Value.TYPE_WARD:
+                district_field = ContactField.get_location_field(self.org, Value.TYPE_DISTRICT)
+                district_value = self.get_field(district_field.key)
+                if district_value:
+                    loc_value = self.org.parse_location(value, 3, district_value.location_value)
```
nicpottier: Let's add some constants for 1,2,3 levels

temba/contacts/models.py
```diff
@@ -371,6 +377,11 @@ def set_field(self, user, key, value, label=None):
             else:
                 loc_value = self.org.parse_location(value, 1)
 
+            if loc_value is not None and len(loc_value) > 0:
+                loc_value = loc_value.first()
+            else:
+                loc_value = None
+
```
nicpottier: Don't understand this block. loc_value should be either None at this point or be a single AdminBoundary no?

temba/flows/models.py
```diff
@@ -5166,7 +5167,7 @@ def evaluate(self, run, sms, context, text):
 
         state = org.parse_location(text, 1)
         if state:
-            return 1, state
+            return 1, state.first()
```
nicpottier: Same here, did the return value for parse_location change and I didn't see it? Isn't it just an admin boundary or None? (on master as far as I can tell parse_location returns one or the other, not a QuerySet)

temba/flows/models.py
```diff
@@ -5197,9 +5199,56 @@ def evaluate(self, run, sms, context, text):
 
         parent = org.parse_location(state, 1)
         if parent:
-            district = org.parse_location(text, 2, parent)
+            district = org.parse_location(text, 2, parent.first())
```
nicpottier: Let's switch 1,2,3 to constants.

temba/contacts/models.py
musamusa: No, I changed the implementation to return a list since we want to query admin boundaries with out going through higher admin levels. that way we can decide on what to do more than one results is returned (example  tofa ward exists in Kano and Sokoto)

temba/flows/models.py
```diff
@@ -5197,9 +5199,56 @@ def evaluate(self, run, sms, context, text):
...
+            if district:
+                return 1, district.first()
+        district = org.parse_location(text, 2)
+
+        #parse location when state contrain is not provide or available
+        if (len(errors) > 0 or state is None) and len(district) == 1:
```
nicpottier: ```if (errors or not State) and len(district) == 1:```

is a lot clearer here

temba/flows/models.py
```diff
@@ -5197,9 +5199,56 @@ def evaluate(self, run, sms, context, text):
...
+
+    def evaluate(self, run, sms, context, text):
+        # if they removed their country since adding the rule
+        org = run.flow.org
+        if not org.country:
+            return 0, None
+        district = None
+
+        # evaluate our district in case it has a replacement variable
+        district_, missing_district = Msg.substitute_variables(self.district, sms.contact, context, org=run.flow.org)
```
nicpottier: No trailing underscores, use ```district_name``` or something else descriptive.

temba/orgs/models.py
```diff
@@ -671,26 +671,44 @@ def parse_decimal(self, decimal_string):
...
+            list.append(text)
+            count -= 1
+
+        return delimiter.join(list)
+
+    def generate_location_query(self, name, level, is_alias=False):
+
+        if is_alias:
+            query = dict(name__iexact=name, boundary__level=level)
+            query[self.repeat_text('parent', level, '__', 'boundary')] = self.country
```
nicpottier: Hrmm. if you really want to get fancy, why not just do:

```'__'.join([parent] * level + ['boundary'])```

File: temba/orgs/models.py [hunk 8]
```diff
@@ -705,26 +723,26 @@ def parse_location(self, location_string, level, parent=None):
         # now look up the boundary by full name
         boundary = self.find_boundary_by_name(location_string, level, parent)
 
-        if not boundary:
+        if len(boundary) == 0:
```
nicpottier: Django QuerySets evaluate to None if they are empty, so these can all still be ```if boundary``` and ```if not boundary```

temba/contacts/models.py
nicpottier: Oh I see it now, I don't see anyplace where that's actually used though.. You always take the first value..

nicpottier:
So this changes the flow definition spec as it introduces a new ruleset type. So we'll need the Java and Python flow engines to be updated as well in this repo (they are pretty simple there since everything is standalone)

Java:
  https://github.com/rapidpro/flows/tree/master/java/src/main/java/io/rapidpro/flows/definition/tests/location

Python:
  https://github.com/rapidpro/flows/blob/master/python/temba_flows/definition/tests.py

musamusa:
@nicpottier PR updated also I'll get back to you when done with the flow definition spec

musamusa:
@nicpottier I'm done with spec definition. the PR is at https://github.com/rapidpro/flows/pull/5

temba/values/models.py
```diff
@@ -464,9 +466,12 @@ def get_value_summary(cls, ruleset=None, contact_field=None, filters=None, segme
                 if not parent_osm_id and field.value_type == Value.TYPE_DISTRICT:
                     raise ValueError(_("You must specify a parent state to segment results by district"))
 
+                if not parent_osm_id and field.value_type == Value.TYPE_WARD:
+                    raise ValueError(_("You must specify a parent state to segment results by ward"))
```
nicpottier: "parent district" not "parent state"

Issue #145, opened 2015-03-21
Add a Vagrantfile for an easier dev environment setup


rodowi:
It would be nice to provide a [Vagrantfile](https://www.vagrantup.com/) to make it easier to load a dev environment with all the tools set up (python version, postgresql, credentials, etc.).

If that makes sense, I'd be happy to open a PR and sign the [contributor agreement](https://github.com/rapidpro/rapidpro/blob/master/CONTRIBUTING.md).

nicpottier:
Hi Rod!  We've never used Vagrant internally before, but that sounds like it would be super useful for people getting started, so would definitely welcome that. If you could write a quick 'howto' to put in the documentation with the usage of said file that'd be awesome.

eduardo-g-silva:
Hi guys any news about the vagrant file?

rodowi:
Sorry guys, I'm no longer working on this.

javaguirre:
Hi! I have a working version with Vagrant, I'll make a PR if you people are interested. :-)

eduardo-g-silva:
I am. Thank s

ewheeler:
added by @javaguirre in https://github.com/rapidpro/rapidpro/pull/254

PR #254, opened 2016-05-10
Vagrant developer environment

javaguirre:
Vagrant developer environment. #145 

Only two dependencies, Vagrant and Ansible.

I use Ansible to configure and manage the developer environment because the configuration is cleaner and often self-explanatory. :-)

The developer environment would work as follows.

```bash
cd rapidpro/ansible
ansible-galaxy install -r requirements.yml
cd ..
vagrant up --provision
```

When It finishes:

```bash
vagrant ssh
cd rapidpro
source .venv/bin/activate
python manage.py runserver 0.0.0.0:8000
```

Then if you go in your browser to `http://localhost:8000` you're good to go!

I'll update the development gh-page configuration if you're ok with this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -108,3 +108,8 @@ docs/_build/
 # PyBuilder
 target/
 textit
+
+# Vagrant/Ansible
+.vagrant/
+ansible/vendor/
+ansible/site.retry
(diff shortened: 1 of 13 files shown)


nicpottier:
This looks good though we have no experience with either of these so can't really speak to whether they are configured right. But definitely willing to merge it. Can you fill out and send the contributors agreement here?

javaguirre:
Done!

nicpottier:
Awesome, thanks!

PR #244, opened 2016-06-13
Custom URL and Version String for Twilio Rest API

dwilkie:
This feature will allow the user to set a custom base URL and Version String when configuring a Twilio channel. The proposed UI would add two optional extra fields to specify the Twilio Rest API Base URL to something other than https://api.twilio.com (default) and specify the version string to something other than "2010-04-01" (default). See image below:

image

If the URL or Version String is specified the TwilioClient would be initialized with the URL as the third argument and the Version String as the forth argument. See: https://github.com/twilio/twilio-python/blob/master/twilio/rest/client.py#L44 and https://github.com/rapidpro/rapidpro/blob/master/temba/channels/models.py#L2730

The purpose of this is to allow users to replace Twilio with something like adhearsion-twilio https://github.com/dwilkie/adhearsion-twilio which they can connect to a local telco.

A couple of initial questions:

1. For the UI should we have an "Advanced Settings" Tab on the Modal dialog for configuring a Twilio Channel (as shown in the diagram above)? Or should we put these two fields directly below the field for the Twilio Account Token and default to their default values?

2. If a custom URL is set then RapidPro should not try to setup the endpoints automatically inside Twilio. This will have to be done manually inside adhearsion-twilio. Where is the code that does this?

3. Are there any integration tests / javascript unit tests to test the UI for configuring a Twilio channel? And are there any unit tests for testing that the Account SID and Token are set and used correctly?


dwilkie:
~ WIP

Comments welcome...

temba/channels/models.py
```diff
@@ -57,6 +57,7 @@
 SHAQODOON = 'SQ'
 SMSCENTRAL = 'SC'
 TWILIO = 'T'
+TWILIO_COMPLIANT_API = 'TCA'
```
nicpottier: This is a bit of a mouthful, how about we use `TWIML_API` everywhere?

temba/orgs/models.py
```diff
@@ -86,6 +86,9 @@
 ACCOUNT_SID = 'ACCOUNT_SID'
 ACCOUNT_TOKEN = 'ACCOUNT_TOKEN'
 
+TWILIO_COMPLIANT_API_ACCOUNT_SID = 'TWILIO_COMPLIANT_API_ACCOUNT_SID'
+TWILIO_COMPLIANT_API_ACCOUNT_TOKEN = 'TWILIO_COMPLIANT_API_ACCOUNT_TOKEN'
```
nicpottier: No need to create new ones, you can use `ACCOUNT_SID` and `ACCOUNT_TOKEN` above.

temba/orgs/models.py
```diff
@@ -594,6 +597,24 @@ def connect_twilio(self, account_sid, account_token):
         self.save(update_fields=['config'])
         self.clear_channel_caches()
 
+    def connect_twilio_compliant_api(self, account_sid, account_token):
+        app_name = "%s/%d" % (settings.TEMBA_HOST.lower(), self.pk)
+        app_url = "https://" + settings.TEMBA_HOST + "%s" % reverse('handlers.twilio_handler')
+
```
nicpottier: ```app_url = "https://%s%s" % (settings.TEMBA_HOST, reverse('handlers.twilio_handler'))```

temba/orgs/views.py
```diff
@@ -425,7 +425,7 @@ def get_object(self, *args, **kwargs):
 class OrgCRUDL(SmartCRUDL):
     actions = ('signup', 'home', 'webhook', 'edit', 'join', 'grant', 'create_login', 'choose',
                'manage_accounts', 'manage', 'update', 'country', 'languages', 'clear_cache', 'download',
-               'twilio_connect', 'twilio_account', 'nexmo_configuration', 'nexmo_account', 'nexmo_connect', 'export',
+               'twilio_connect', 'twilio_compliant_api_connect', 'twilio_account', 'twilio_compliant_api_account', 'nexmo_configuration', 'nexmo_account', 'nexmo_connect', 'export',
```
nicpottier: same here `twiml_api_connect` etc..

temba/orgs/views.py
```diff
@@ -1489,6 +1489,7 @@ def derive_formax_sections(self, formax, context):
                 client = org.get_twilio_client()
                 if client:
                     formax.add_section('twilio', reverse('orgs.org_twilio_account'), icon='icon-channel-twilio')
+                formax.add_section('twilio-compliant-api-channel', reverse('orgs.org_twilio_compliant_api_account'), icon='icon-channel-twilio')
```
nicpottier: This should only appear if it is already connected.

templates/channels/channel_claim.haml
```diff
@@ -252,6 +252,22 @@
...
+                .claim-glyph.icon-channel-twilio
+                %p.claim-text
+                  -trans "Connect Twilio Compliant API"
+
+          .span9
+            .claim-description
+              %div
+                -blocktrans
+                  Connect to a Twilio Compliant API such as <a href=https://github.com/dwilkie/adhearsion-twilio>adhearsion-twilio</a>. You can use this to migrate away from Twilio and save on Twilio fees.
+
```
nicpottier: Don't link to any provider here, just say Twilio Compliant API. No need for saving fees text etc..

nicpottier:
Overall looks good. As I said, let's move away from the mouthful of Twilio Compliant API to something more like TWIML API. 

Had a few other random comments, we'll need tests on all this before we can merge, but otherwise looks good.

temba/orgs/models.py
dwilkie: I think we still need these (although I renamed them to `TWIML_API_ACCOUNT_SID` and `TWIML_ACCOUNT_TOKEN` as per your suggested rename). Reason being these keys are stored in the `orgs_org#config column`. If you use `ACCOUNT_SID` and `ACCOUNT_TOKEN` then you can't connect a Twilio Account and a TwiML at the same time.

nicpottier: You are right about this, but I think it actually points out that maybe we should be using a slightly different approach. Twilio and Nexmo being org scoped is actually a regret of ours in the platform, it causes some issues in various places.

Especially since this channel type has different possible endpoints (while Twilio always goes to the same place) this configuration does belong on the channel objects with its own endpoint as opposed to being org level. (ie, more similar to the Kannel endpoint than the current Twilio endpoint)

What do you think?

dwilkie: Sure totally agree. I'll have a look how the Kannel endpoint is setup and try to adapt this channel from there

nicpottier: Out of curiosity, how far does the Twilio compatibility go? Does it include searching for available numbers for example? If not, then that definitely is another argument towards a more Kannel like approach, if it does, then that may indeed be more similar to the current Twilio approach.

dwilkie: No there's no searching for available numbers or anything. In fact the Twilio REST Client doesn't need to be used at all for setting up the channel. The webhook URL https://github.com/dwilkie/rapidpro/blob/55487f22f0915dad4dfb4391d72c198dd547d65b/temba/orgs/models.py#L616 needs to be displayed somewhere in the channel configuration and set up manually inside Adhearsion

dwilkie:
@nicpottier I updated the PR with your suggestions. I.e. moved the channel configuration out of the org and into the channel. I got it to try and make an outgoing call to the URL specified in the configuration. I still need to get the tests to pass here, but do you see anything else that's needed?

templates/channels/channel_configuration.haml
```diff
@@ -639,6 +639,20 @@
...
+        To finish configuring your connection you'll need to set the following status callback URL in your Adhearsion Application
+
+    %hr
+
+    %h4
+      -trans "Status Callback URL"
+
+    %code
+      https://{{ domain }}{% url 'handlers.twilio_handler' %}
+
```
nicpottier: Looks like you potentially create API tokens as well when adding a new channel, seems those should be displayed here.

nicpottier:
Hi David. Yep, this looks sane, definitely interested to see this running.. have you played with it yet and had calls working end to end?

Out of curiosity is it possible to get Adhersion running on a RaspberryPi with a GSM modem for voice?

This is an aside, but we've been mulling over different possibilities in getting local IVR, and one thought has been to integrate the standalone RapidPro flow engine (https://github.com/rapidpro/flows) into a Raspberry Pi appliance that could run a call completely locally then ship results back up. Wonder if Adhersion would be the right glue for us to use between Asterisk and our engine or whether writing something that went straight to Asterisk makes more sense.

dwilkie:
Hi Nic. Yeh still got some work to do, but i've got RapidPro to use the new channel to send outbound calls a local REST API server. The plan is to have that REST API Server (which talks Twilio's REST API) to enqueue calls via Sidekiq. Then a sidekiq worker instance will read the queue and use DRb to enqueue outbound calls in Adhearsion. Adhearsion then talks to FreeSwitch to make the outbound call via SIP.

I believe that you'll have more luck running FreeSwitch on a Raspberry Pi than Asterisk. For example you can run FreeSwitch on a router running OpenWrt quite easily.

Adhearsion is great in that you can plug it into both FreeSwitch and Asterisk, but the issue is that it's written in Ruby so I don't believe you'll be able to get that running on a Rasberry Pi. You might have better luck writing something straight for FreeSwitch. Take a look at FreeSwitch's httapi module.

Happy to help you out with this. Feel free to move the conversation into an Email.

smn:
For what it's worth, we've got a [freeswitch channel](http://github.com/praekelt/vumi-freeswitch-esl/) that can be launched with [junebug](http://junebug.readthedocs.io/en/latest/). The channel talks over the event socket to Freeswitch and does things like accepting inbound calls, relay DTMF signals and initiate outbound calls. Just posting it here in case it could be of use more widely.

dwilkie:
Closing in favor of #317

PR #260, opened 2016-06-13
add config fields to kannel update form

aristide:
While testing Kannel i needed to update config fields. So i added some code i wanted to share


nicpottier:
We actually make Channels immutable on purpose, as it is makes sure there isn't any confusion as to how a message was sent. The way to update a channel is to remove it, then re-add it.

Issue #277, opened 2016-07-18
Add Flow base language to flow events webhook 


xkmato:
The need for base language to be exposed. This is useful when trying to predict what keys to expect in the `categories dict` in values

nicpottier:
Isn't this part of the flow definition itself?

ewheeler:
yes. currently, flow webhook consumers would have to get the flow definition to know what language keys to expect for `category`

it would also be nice to expose which language rapidpro has served to the flowrun's contact

nicpottier:
We don't actually record which language a contact went through at any particular step (it would have to be per step not per run)

I'd rather have clients do a single hit to grab the definition to know the default language than add a new field to every step in the API.

ewheeler:
Ok makes sense for which lang a contact went through.

Currently there's no need to have API access to consume flow events, so to properly consume the event we'll have to issue tokens to consumers in order to prefetch the flow def to find out the base language.

nicpottier:
Sorry, I was misunderstood what was being asked for here, thought you guys were talking about the events on the runs endpoint.

What exactly are you guys trying to do? I do agree that having the category be a map is actually kinda annoying and maybe that should just be the base language category, that's a backwards breaking change which isn't great though. Also think having the contact langauge at that point in time wouldn't be the worst thing in the world, that's an easy one at least.

ewheeler:
ahh ok. talking about flow event webhooks here.

we're working on a little webservice to update a google fusion table with flow results. add a webhook to the end of the flow, then this service adds as a row for each run (and even creates the fusion table if one doesnt exist for the flow yet!). 
since this could be used across orgs with different base langs, we'd like to use the `category` for multiple choice steps in the org's base lang when adding to the fusion table. without knowing the org's base lang you'd have to guess

Issue #266, opened 2016-07-05
Add Flow Name to flow events webhook


xkmato:
Adding the flow name to the flow events webhook would be very helpful for use with webhooks that are designed to work more generically on multiple flows.

nicpottier:
Good suggestion. We'd accept a PR that added a `flow_name` parameter to the webhook payload.

PR #278, opened 2016-07-19
Expose flow name and baselanguage in webhook flow events

xkmato:
As a solution for #277 and #266 

* Add `flow_name`
* Add `flow_base_language`

The point of adding base language it t make it easier(possible) to predict the keys(or one of the keys) of the `category` dictionary in vaules


nicpottier:
Yep.

PR #321, opened 2016-08-10
Feature/ui improvements fb channel

nicpottier:


nicpottier:
So this isn't an approach that could scale, Msg sending has to be the highest throughput portion of RapidPro and hitting the DB to try to derive rules and how they might be mapped to buttons really wouldn't work.

The better approach here is to change the flow editing UI to allow for the creation of the various FB items like menus etc.. and then allow branching around them. The menu itself would have to be encoded in the Msg body.

Just brainstorming but one option there would be to use the "media" attribute on Msg and add options for say "fb:menu" or something and then render appropriately. Would have to think through that but the sum total is that the Msg itself needs to encode the menu etc.. you can't create a Msg of one thing and then send something else as you have it here.

johncordeiro:
I agree with you, the best solution for facebook guidelines is creating a different UI for enable the options available on Facebook Messenger.

Today we have some kind of flows: `Messaging`, `Phone call` and `Surveyor`. What do you think to create a different kind for `Facebook` here:

Facebook Flow

There will be the facebook guideline options here:

Ruleset

What do you think about this option?

nicpottier:
So I don't think this is the right direction either, this probably requires something quite a bit more complicated to do right.

If we treated Facebook flows as totally separate, or even just IP based messaging flows as totally separate, that would mean say U-Report would have to run multiple flows for a survey and somehow merge the results on the backend etc.. with all the problems associated with them having the same names for responses etc.. They would also need to track the respondents in groups associated with their channel type, which could turn into a mess if people have more than one URN etc..

This is really probably something that is more like localization in a way. In that there would be a switch on a flow when an org has IP based channels that would let you edit the text for responses and rulesets etc, to allow for rich menu based responses. The rulesets in the end would be the same but would have a different 'view' for those cases and flow authors would be able to switch not only between languages when editing but between 'rich' and 'plain' messaging authoring or the like. Which then comes with a multiplication of options in how the flow specification works etc..

So ya, don't see any easy way to do this correctly, would be a ton of work that would require a lot of thought in the right approach.

nicpottier:
We chatted a bit more about this and actually have any idea that might be more reasonable.

So the big change here is really outbound menus and buttons and such, waiting for rich media like photos and videos should be standard options for text etc, with a "buyer beware" attitude that people doing this on text only SMS channels won't have much luck.

The idea of having buttons or menus for FB and Telegram (or possibly other channels in the future) is more specific though. I think what we can do here is have a new ruleset type similar to what is being done on the USSD side that encapsulates both the question and the ruleset answers. We can then render this differently for text and IP based channels that support menus or buttons. That ruleset will have to include the question text for normal text based channels and a different set of text for button based channels. (as the former will require the user to respond with a keyword, number or letter, while the former they can just click on something)

But that avoids the royal mess of having a separate rendering per channel.

nicpottier:
Going to close this since the approach is going to have to be quite a bit different, but definitely love the product!

nicpottier:
We're thick in a rewrite of the engine and editor right now, but both of those will enable this type of thing much more easily than the current architecture.

PR #334, opened 2016-08-30
fix migration dependencies/ordering

nullcode:
Running migration flows.60 throws this:
"django.db.utils.ProgrammingError: column flows_flowrun.timeout_on does not exist"
That column was only added in 61, so I fixed the order of dependencies


nicpottier:
ooh, right you are, thanks!

Issue #298, opened 2016-07-26
Add a LINE channel


ycleptkellan:
https://developers.line.me/bot-api/api-reference

nicpottier:
Added.

PR #357, opened 2016-09-26
LINE channel

teehamaral:
This PR includes LINE channel referenced in #298


temba/channels/handlers.py
```diff
@@ -1954,3 +1954,40 @@ def post(self, request, *args, **kwargs):
 
         else:  # pragma: no cover
             return HttpResponse("Not handled, unknown action", status=400)
+
+
+class LINEHandler(View):
```
ericnewcomer: While presenting to the user it's okay to use LINE since that's their brand, but lets stick with Line instead of LINE in code.

temba/channels/handlers.py
```diff
@@ -1954,3 +1954,40 @@ def post(self, request, *args, **kwargs):
...
+
+            for r in receive:
+                data = r._Message__data
+                content = data.get('content')
+                text = content.content.get('text')
+                date = ms_to_datetime(data.get('created_time'))
+                from_mid = data.get('from_mid').encode('utf-8')
+                Msg.create_incoming(channel=channel, urn=URN.from_line(from_mid), text=text, date=date)
+
+            return HttpResponse("SMS Accepted")
```
ericnewcomer: SMS?

temba/channels/migrations/0039_auto_20160926_1316.py
```diff
@@ -0,0 +1,19 @@
...
+        ('channels', '0038_merge'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='channel',
+            name='channel_type',
+            field=models.CharField(default='A', help_text='Type of this channel, whether Android, Twilio or SMSC', max_length=3, verbose_name='Channel Type', choices=[('AT', "Africa's Talking"), ('A', 'Android'), ('BM', 'Blackmyna'), ('CT', 'Clickatell'), ('EX', 'External'), ('FB', 'Facebook'), ('GL', 'Globe Labs'), ('HX', 'High Connection'), ('H9', 'Hub9'), ('IB', 'Infobip'), ('JS', 'Jasmin'), ('KN', 'Kannel'), ('LN', 'LINE'), ('M3', 'M3 Tech'), ('MB', 'Mblox'), ('NX', 'Nexmo'), ('PL', 'Plivo'), ('SQ', 'Shaqodoon'), ('SC', 'SMSCentral'), ('ST', 'Start Mobile'), ('TG', 'Telegram'), ('T', 'Twilio'), ('TMS', 'Twilio Messaging Service'), ('TT', 'Twitter'), ('VB', 'Verboice'), ('VI', 'Viber'), ('VM', 'Vumi'), ('VMU', 'Vumi USSD'), ('YO', 'Yo!'), ('ZV', 'Zenvia')]),
+        ),
+    ]
```
ericnewcomer: This should be a single migration, not 5.

temba/channels/tests.py
```diff
@@ -7485,3 +7510,79 @@ def test_send(self):
             msg.refresh_from_db()
             self.assertEquals(ERRORED, msg.status)
             self.clear_cache()
+
+
+class LINETest(TembaTest):
```
ericnewcomer: Normal case here too

temba/channels/tests.py
```diff
@@ -7485,3 +7510,79 @@ def test_send(self):
...
+        self.assertEquals("123456789", msg.contact.get_urn(LINE_SCHEME).path)
+        self.assertEquals(self.org, msg.org)
+        self.assertEquals(self.channel, msg.channel)
+        self.assertEquals("Hello, BOT API Server!", msg.text)
+
+    def test_send(self):
+        joe = self.create_contact("Joe", urn="line:1234")
+        msg = joe.send("Hello, BOT API Server!", self.admin, trigger_send=False)
+
+        settings.SEND_MESSAGES = True
```
ericnewcomer: This is a no no.

temba/channels/views.py
```diff
@@ -1844,6 +1846,59 @@ def form_valid(self, form):
 
             return HttpResponseRedirect(reverse('channels.channel_configuration', args=[channel.id]))
 
+    class ClaimLine(OrgPermsMixin, SmartFormView):
+        class LineForm(forms.Form):
+            channel_id = forms.CharField(label=_("Channel ID"), required=True, help_text=_("The ID of the your Channel on Business LINE"))
+            channel_secret = forms.CharField(label=_("Channel Secret"), required=True, help_text=_("The Secret of the your Channel on Business LINE"))
+            channel_mid = forms.CharField(label=_("Channel MID"), required=True, help_text=_("The MID of the your Channel on Business LINE"))
```
ericnewcomer: Do they call them Channel's at LINE? If that's the case, then make these read something like:

"The MID of your Business LINE Channel"

Either way, "of the your" on all of these is not right.

temba/channels/views.py
```diff
@@ -1844,6 +1846,59 @@ def form_valid(self, form):
...
+                from linebot.client import LineBotClient
+
+                channel_id = self.cleaned_data.get('channel_id')
+                channel_secret = self.cleaned_data.get('channel_secret')
+                channel_mid = self.cleaned_data.get('channel_mid')
+
+                existing = Channel.objects.filter(channel_type=Channel.TYPE_LINE, address=channel_mid, is_active=True).first()
+                if existing:
+                    raise ValidationError(_("Channel with this address already exists, please try again with other."))
+
```
ericnewcomer: Language here isn't quite right.

A Channel with this address already exists, please try again with a different address.

temba/channels/views.py
```diff
@@ -1844,6 +1846,59 @@ def form_valid(self, form):
...
+                try:
+                    users = line_bot_client.get_user_profile(channel_mid)
+                    profile = [
+                        {'mid': user._UserProfile__profile.get('mid'),
+                         'picture_url': user._UserProfile__profile.get('picture_url'),
+                         'display_name': user._UserProfile__profile.get('display_name'),
+                         'status_message': user._UserProfile__profile.get('status_message')} for user in users][0]
+                except:
+                    profile = {'mid': channel_mid, 'picture_url': None, 'display_name': channel_id, 'status_message': None}
+
```
ericnewcomer: Why continue if we fail to get the bot info? This should raise an exception.

temba/contacts/migrations/0043_merge.py
```diff
@@ -0,0 +1,15 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('contacts', '0042_remove_exportcontactstask_host'),
+        ('contacts', '0042_auto_20160830_1727'),
+    ]
+
+    operations = [
+    ]
```
ericnewcomer: why?

templates/channels/channel_claim_line.haml
```diff
@@ -0,0 +1,59 @@
...
+-block pre-form
+  .info
+    %p
+      -blocktrans with name=brand.name
+        You can connect a <a href="https://developers.line.me/bot-api/overview" target="_blank">LINE Bot</a> to your {{ name }} account
+        to automate sending and receiving LINE messages.
+    %p
+      -blocktrans
+        You will need to <a href="https://developers.line.me/type-of-accounts/bot-api-trial" target="_blank">create a new LINE bot</a> and
+        get its Authentication Credentials. To do so:
```
ericnewcomer: After you <a>create a LINE bot</a>, you can use its authentication <a>credentials</a> to add it as a Channel.

templates/channels/channel_claim_line.haml
```diff
@@ -0,0 +1,59 @@
...
+            Access the page <a href="https://developers.line.me/channels/" target="_blank">LINE Channels</a> e add a new Channel.</li>
+
+        %li
+          -blocktrans
+            After add, select your Channel, go to basic information and click EDIT.
+
+        %li
+          -blocktrans
+            Enter the callback URL in the Callback URL section (Generated by RapidPRO after this step).
+
```
ericnewcomer: Language is a bit goofy through this whole section. Also make sure to reference {{brand}} instead of RapidPro.

ericnewcomer:
@rowanseymour have a look through this tomorrow if you would, thanks!

temba/channels/models.py
```diff
@@ -528,6 +534,15 @@ def add_facebook_channel(cls, org, user, page_name, page_id, page_access_token):
         return channel
 
     @classmethod
+    def add_line_channel(cls, org, user, credentials, name):
+        channel_id = credentials.get('channel_id')
+        channel_secret = credentials.get('channel_secret')
+        channel_mid = credentials.get('channel_mid')
+
+        channel = Channel.create(org, user, None, Channel.TYPE_LINE, name=name, address=channel_mid, config={Channel.CONFIG_CHANNEL_ID: channel_id, Channel.CONFIG_CHANNEL_SECRET: channel_secret, Channel.CONFIG_CHANNEL_MID: channel_mid})
```
rowanseymour: `return Channel.create...`

temba/channels/tests.py
rowanseymour: Pattern used in some other channel tests is 

```python
with self.settings(SEND_MESSAGES=True):
    ...
```

templates/includes/nav.haml
```diff
@@ -1,4 +1,4 @@
--load smartmin i18n
+-load smartmin i18n compress temba
```
rowanseymour: Not getting why this is here?

temba/channels/views.py
teehamaral: Yes, they call "Channels" at LINE developers console. But, also has reference with "BOT API", so, could be "The MID of the Line bot"?

ericnewcomer: Definitely prefer bot in that case to avoid confusion with rapid channel

teehamaral: So, I'll change "channel" to "bot", ok?

teehamaral:
Review, please.

ericnewcomer:
I went poking around the LINE website to create a bot and noticed that this uses an API that is already deprecated, no?
https://developers.line.me/bot-api/overview

We should look at using the current api instead:
https://devdocs.line.me/en/#messaging-api

teehamaral:
Oh yeah! It's true. I'll change to the current api.

teehamaral:
Review, please.

temba/channels/handlers.py
```diff
@@ -1954,3 +1954,42 @@ def post(self, request, *args, **kwargs):
...
+        from temba.msgs.models import Msg
+
+        channel_uuid = kwargs['uuid']
+
+        channel = Channel.objects.filter(uuid=channel_uuid, is_active=True, channel_type=Channel.TYPE_LINE).exclude(org=None).first()
+        if not channel:
+            return HttpResponse("Channel with uuid: %s not found." % channel_uuid, status=404)
+
+        try:
+            data = request.body.decode('utf-8', errors='ignore')
```
nicpottier: Is this needed? Django should be properly decoding the body of the request based on the headers passed.

temba/channels/handlers.py
```diff
@@ -1954,3 +1954,42 @@ def post(self, request, *args, **kwargs):
...
+                source = item.get('source')
+                message = item.get('message')
+
+                if message.get('type') == 'text':
+                    text = message.get('text')
+                    user_id = source.get('userId')
+                    date = ms_to_datetime(item.get('timestamp'))
+                    Msg.create_incoming(channel=channel, urn=URN.from_line(user_id), text=text, date=date)
+
+            return HttpResponse("Msg Accepted")
```
nicpottier: Shouldn't return this in the case where we don't do anything (IE, the type is not 'text'). Perhaps "Msg Ignored" in that case?

temba/channels/handlers.py
```diff
@@ -1954,3 +1954,42 @@ def post(self, request, *args, **kwargs):
...
+        try:
+            data = request.body.decode('utf-8', errors='ignore')
+            content = json.loads(data)
+            events = content.get('events')
+
+            for item in events:
+                source = item.get('source')
+                message = item.get('message')
+
+                if message.get('type') == 'text':
```
nicpottier: Should validate and fail with 400 if source, message or type are not in the request.

temba/channels/models.py
```diff
@@ -1143,6 +1158,32 @@ def send_facebook_message(cls, channel, msg, text):
...
+        from temba.msgs.models import Msg, WIRED
+
+        channel_access_token = channel.config.get(Channel.CONFIG_AUTH_TOKEN)
+
+        data = json.dumps({'to': msg.urn_path, 'messages': [{'type': 'text', 'text': text}]})
+
+        start = time.time()
+        response = requests.post('https://api.line.me/v2/bot/message/push',
+                                 data=data,
+                                 headers={'Content-Type': 'application/json', 'Authorization': 'Bearer %s' % channel_access_token})
```
nicpottier: Should be using TEMBA_HEADERS + these

temba/channels/tests.py
```diff
@@ -4563,7 +4600,7 @@ def test_received(self):
 
         # load our message
         msg = Msg.objects.get()
-        self.assertEquals('+977788123123', msg.contact.get_urn(TEL_SCHEME).path)
+        self.assertEquals('977788123123', msg.contact.get_urn(TEL_SCHEME).path)
```
nicpottier: Why is this different?

temba/channels/tests.py
```diff
@@ -4696,7 +4733,7 @@ def test_received(self):
...
```
nicpottier: Same here, why is this different?

templates/channels/channel_claim.haml
```diff
@@ -646,6 +646,24 @@
                     and receive messages.
 
 
+      -if org_perms.channels.channel_claim_line
+        .claim-row.twitter-channel
```
nicpottier: wrong class

temba/channels/tests.py
ericnewcomer: This is due to the phonenumbers library updating which is now more accurate in rejecting nepalese numbers. This is not the correct fix, if you merge the master branch into yours, you'll get the proper fix for this.

teehamaral: Ok.

teehamaral:
Review, please.

temba/channels/models.py
```diff
@@ -1143,6 +1158,34 @@ def send_facebook_message(cls, channel, msg, text):
...
+            Msg.mark_sent(channel.config['r'], channel, msg, WIRED, time.time() - start, external_id=None)
+
+            ChannelLog.log_success(msg=msg,
+                                   description="Successfully delivered",
+                                   method='POST',
+                                   url=response.request.url,
+                                   response=response.content,
+                                   response_status=response.status_code)
+        else:
+            ChannelLog.log_error(msg=msg, description=content.get('message'))
```
nicpottier: When logging errors the pattern is generally to put all the information possible to help in debugging. IE, the URL, response, content etc.. (this should be in the log_success) as well

templates/channels/channel_configuration.haml
```diff
@@ -598,6 +598,18 @@
       https://{{ domain }}{% url 'handlers.viber_handler' 'receive' object.uuid %}
 
 
+  -elif object.channel_type == 'LN'
+    %h4
+      -blocktrans
+        To finish configuring your Line channel you'll need to set the following callback URL on your LINE Channel
```
nicpottier: inconsistent case

teehamaral: Could you to say how would be better?

temba/channels/models.py
teehamaral: OK

templates/channels/channel_configuration.haml
nicpottier: You use Line and LINE in the same sentence. Just pick one.

nicpottier: sure

nicpottier:
I tried hooking up an actual Line channel and it didn't work because the channel mid is longer than 16 characters and Channel.address is limited to that. So you'll need to also add a migration to lengthen the max length of address.

Did you actually test this? Kind of surprised I ran into that..

teehamaral:
Sorry, I tested, but I think I made a mistake in the process removing unnecessary migrations. The changes already were pushed.

nicpottier:
On the line config page (where we give the callback URL) we should also give the IP addresses to whitelist. (there is a config settings that can be used here, look at Viber for an example)

teehamaral:
Yes, I've pushed the change.

nicpottier:
So I've been walking through hooking one of these up and the various steps you have to go through is super annoying and hard to find. I can't imagine users are going to be able to do this on their own without significant hand holding.

So let's add a step by step guide that helps users:

1) Creating a new Line@ account
2) Walks them through enabling the developer account for chat
3) Walks them through getting setting the various parameters for their bot (IE, here's how to find your secret and access token, here's where to set webhook URL and whitelist IPs)
4) Finally, shows them how they can add the bot to test it and advises them on getting final approval on their bot

I think from a technical standpoint this PR looks ok but because Line is so shitty about their docs we need to provide our own docs for all the above for anybody to actually have any success with it.

teehamaral:
Absolutely. I also think the Line documentation to add bot is horrible. I will try to do a step by step to improve the way the addition of a new Line channel.

teehamaral:
Do you think we could do like how the Android channel tutorial?

nicpottier:
I think something like that could work, though would probably warn against using screenshots, but instead describing the actions, just because I imagine stuff will change and it is a lot harder to update screenshots than just text.

teehamaral:
Yeah, alright.

johncordeiro:
Do you think that modifications built by @teehamaral attend your changes requests?

templates/channels/channel_claim_line.haml
```diff
@@ -0,0 +1,117 @@
...
+          -blocktrans
+            Access the <a href="https://developers.line.me/channels/" target="_blank">LINE Channels</a> page and add a new Channel.</li>
+
+        %li
+          -blocktrans
+            Create a Line account using the application on your smartphone
+
+        %li
+          -blocktrans
+            Active the permission to login via web application in the "Settings > Account > Allow login"
```
nicpottier: `Activate the permission`...

templates/channels/channel_claim_line.haml
```diff
@@ -0,0 +1,117 @@
...
+                Select "Start using Messaging API"
+
+            %li
+              -blocktrans
+                Enter your credentials created through the application and wait for the confirmation code in it.
+
+              .important
+                %p
+                  -blocktrans
+                    Note: At this time will be create a LINE business account and some information, for example, address, phone, etc., will be requested.
```
nicpottier: `At this time you must create a LINE business account..`

templates/channels/channel_configuration.haml
```diff
@@ -627,6 +627,38 @@
...
+      %p
+
+        %ol.line-steps
+          %li
+            -blocktrans
+              Configure "Callback URL" in the channel page (the same page which get the information Channel Secret and Channel Access Token) by clicking on the "Edit" button, filling the field "webhook URL" and pressing on the "Save" button.
+
+          %li
+            -blocktrans
+              Enter also the IP addresses in the "Server IP Whitelist" menu with the list of addresses that will be displayed after the correct filling out the form below.
```
nicpottier: `Fill the IP addresses in the "Server IP Whitelist" with the list of addresses displayed below.`

nicpottier:
Thanks guys!

PR #360, opened 2016-10-04
added omitted template library load statement in templates/includes/n…

antsmc2:
Just added omitted load statement for temba in templates/includes/nav.haml.

A page using this template file threw exceptions until the library was loaded.


ericnewcomer:
As a formality, we need this before we can merge:

Seems overkill for just this PR, but at least you should be good going forward.

nicpottier:
Fixed in another PR.

PR #368, opened 2016-10-18
USSD Pull with new Trigger type + VUMI USSD Pull support

nullcode:
Hi!
This PR includes the new trigger type to implement USSD Pull. It's following the same principles as other trigger types.
One important note on the implementation since it might affect other trigger types too: I migrated the channel assigned to a trigger from 1-1 to 1-Many because in this case it makes sense to assign more ussd codes (keywords) to the same channel since Vumi Go routing interface allows that.

How it works in practice ?
1. Add a VUMI USSD channel
2. Add a USSD mobile initiated flow trigger
3. Call the designated ussd code from a handset and receive the first menu from the triggered flow

screen shot 2016-10-18 at 22 55 04


nullcode:
@ericnewcomer @nicpottier can you guys take a look at this ?
also I have a test failing on this branch, but it seems unrelated to my stuff, do you have any idea what that can be?

`ERROR: test_new_conversation_trigger (temba.triggers.tests.TriggerTest)`


`Exception: Unable to update call to action: {"error":{"message":"Invalid OAuth access token.","type":"OAuthException","code":190,"fbtrace_id":"Aort2jn8cUR"}}
`

nicpottier:
Hey @nullcode, so sorry for the late response, please feel free to prod us if we don't respond within a day or two.

Have you pulled in master? Master seems to be passing so I wonder if you branched at a time when it was broken perhaps? That error to me looks like we are trying to call out to Facebook which shouldn't be happening at all, so not sure what is awry.

nullcode:
@nicpottier It was pretty hard to find out, because it was from the test that does a request that should fail with a form error, but doesn't in my case, because I changed the model and that call being not mocked threw the error.
So, I believe we should first discuss if you agree with the 1toMany relationship with trigger.channel model change and then we can figure out how to deal with "duplicates" where it doesn't make sense.

nicpottier:
Definitely cool with the foreign key change, that definitely makes sense. There may even be an argument for that to be a many to many field (allowing the user to select more than one channel a trigger applies to). We've actually been discussing having a topic for the hackathon be to allow channel selection on the other trigger types.

The bigger concern I have here is on the new msg states. These really shouldn't be representing message types (interrupted, triggered), as they are used to represent the state of the message within the system, not what type of message they are. There are a lot of moving pieces that make sure that things work with the defined states so we don't want to muddy the meaning there.

We may need to treat this a bit more like IVR calls which also are session based and therefore have some associated state which is independent of the msg itself. You are going to be at the conference next week yes? Would be great to chat about this in person.

nullcode:
States are indeed a bit tricky there, I could have just introduced another field to deal with these special types of messages, although this was just way less code... anyway looking forward to talk in person. I will bring another nice topic to talk about, namely the USSD shortcodes (eg: *111#1#2#4#). I have a solid implementation now, but would like to discuss it further with you guys. Also excited about all the other developments and talks regarding RapidPro, see you there.

nicpottier:
Ya, thinking about this a bit more I do think the right call here is probably something closer to the IVRCall implementation, but definitely worth a discussion next week with @ericnewcomer. As Msg is the model we have the most of (hundreds of millions of rows) we are pretty sensitive about adding new fields to it. We've found that picking a solution based on less code often comes back to bite you hard. :)

PR #381, opened 2016-11-16
InfoBip USSD Channel WIP

nullcode:
This PR includes the implementation of the integration of InfoBip USSD API as a new USSD channel.

@ericnewcomer @nicpottier The API is different to what we are used to. It's one directional, meaning there is only one way of communication: gateway (API) --> RapidPro. Since the handling of an incoming message and the reply to that message are handled in an async fashion in RapidPro, I've faced some difficulties of making an implementation that is similar to what we did in the past. I'll describe it with an example:
1. The first call is the initiation (will call a `<session_id>/start` endpoint) that includes the msisdn and the shortcode (starcode) the request was initiated with.
2. The response to a call should include the reply message (`ussdMenu`) and if we should close the call.
In RapidPro I can easily process the incoming requests, but the response should include the menu created which is handled asynchronously in the `Msg.process_message` -> `Flow.find_and_handle` -> `Flow.handle_destination` which ends up in a `run.flow.org.trigger_send(msgs)` call that actually sends out the messages. Now here I'd need to make all this chess game synchronously, so I can generate the reply message in the same function that handles the incoming requests.

Can you recommend a best practice approach here? I could potentially call find_and_handle directly and generate the reply (I would need to make some changes on it too), but this could result in some unwanted behaviour that I couldn't foresee.


nullcode:
@nicpottier @ericnewcomer can you please take a look at this first, otherwise I don't have anything to progress with.

nicpottier:
This should be similar to how IVR is handled, in that we also need to process the flow synchronously when Twilio calls into us and return what should happen next. So I would look there for inspiration.

Another reason/way that USSD and IVR are similar, so that actually makes me feel good about the direction. (look at temba/ivr/views/CallHandler

smn:
For what it's worth, Junebug can expose [Vumi's Infobip transport](https://github.com/praekelt/vumi/blob/develop/vumi/transports/infobip/infobip.py) as a channel in RapidPro. For people wanting Infobip in the short term, that may be an option.

PR #394, opened 2016-12-06
Firebase Cloud Messaging channel

teehamaral:
This PR includes Firebase Cloud Messaging channel.

The way that we are saving URN path is different from other channels, to avoid changes in FCM token and duplicated contacts. We are using FCM Token as path and custom id as URN. This is the first idea without changing contact fields too much.


nicpottier:
Woot, excited to see this. Will take a closer look at this tomorrow, but any reason the two parts can't be encoded in the URN?

IE: `fcm:[part1]@[part2]` or the like?

temba/channels/models.py
```diff
@@ -1105,6 +1129,48 @@ def success(cls, channel, msg, msg_status, start, request_method=None, request_u
...
+        if api_key:
+            url = 'https://fcm.googleapis.com/fcm/send'
+            title = channel.config.get(Channel.CONFIG_FCM_TITLE)
+            data_message = {'type': 'rapidpro', 'title': title, 'message': text, 'message_id': msg.id}
+            payload = json.dumps({
+                'data': data_message,
+                'content_available': True,
+                'to': msg.urn_path,
+                'notification': {
+                    "title": title,
```
nicpottier: single quote

temba/channels/models.py
```diff
@@ -1105,6 +1129,48 @@ def success(cls, channel, msg, msg_status, start, request_method=None, request_u
                                   request_time=request_time)
 
     @classmethod
+    def send_fcm_message(cls, channel, msg, text):
+        from temba.msgs.models import Msg, WIRED
+        start = time.time()
+
+        api_key = channel.config.get(Channel.CONFIG_FCM_KEY) if Channel.CONFIG_FCM_KEY in channel.config else None
```
nicpottier: Is there any case where API key would be missing?

temba/channels/tests.py
```diff
@@ -8187,3 +8212,74 @@ def test_send(self):
...
+                mock.return_value = MockResponse(200, '{ "success": 1, "multicast_id": 123456, "error": 0 }')
+
+                # manually send it off
+                Channel.send_message(dict_to_struct('MsgStruct', msg.as_task_json()))
+
+                # check the status of the message is now sent
+                msg.refresh_from_db()
+                self.assertEqual(msg.status, WIRED)
+                self.assertTrue(msg.sent_on)
+                self.clear_cache()
```
nicpottier: Good to check that the post was made with the arguments you expect. (see `mock.assert_called_with`)

temba/channels/views.py
```diff
@@ -1883,6 +1884,32 @@ def get_context_data(self, **kwargs):
...
+        form_class = ClaimFcmForm
+        fields = ('title', 'key',)
+        title = _("Connect Firebase Cloud Messaging")
+        success_url = "id@channels.channel_configuration"
+
+        def form_valid(self, form):
+            org = self.request.user.get_org()
+
+            if not org:  # pragma: no cover
+                raise Exception(_("No org for this user, cannot claim"))
```
nicpottier: These are spurious, there's no way we could get here without an org (because of permissions), we should remove all of these.

templates/channels/channel_claim.haml
```diff
@@ -679,6 +678,25 @@
...
+          %a.btn.btn-primary.btn-primary-muted.btn-claim.connect-gcm{href:"{% url 'channels.channel_claim_fcm' %}"}
+            .claim-glyph.icon-fcm
+            %p.claim-text
+              -trans "Firebase Cloud Messaging"
+
+        .span9
+          .claim-description
+            %div
+              -blocktrans
+                Add a Firebase Cloud Messaging Channel to send and receive messages.
```
nicpottier: Add a link here to the google Firebase page.

templates/channels/channel_configuration.haml
```diff
@@ -432,6 +432,51 @@
...
+        .value {{ object.config_json.FCM_KEY }}
+
+    .clearfix
+    %hr
+
+    %h4
+      -trans "Contact Register"
+
+    %p
+      -trans "To register contacts, is recommended that you send the data to URL below on your App."
```
nicpottier: To register contacts, POST to the following URL with the parameters `urn`, `fcm_token` and optionally `name`

templates/channels/channel_configuration.haml
```diff
@@ -432,6 +432,51 @@
...
+
+    %code
+      https://{{ domain }}{% url 'handlers.fcm_handler' 'register' object.uuid %}
+
+    %h4
+      -trans "Receive URL"
+
+    %p
+      -trans "To receive incoming messages, you need to send the data to URL below on your App."
```
nicpottier: To handle incoming messages, post to the following URL with the parameters `from`, `msg` and `fcm_token`.

templates/channels/channel_configuration.haml
```diff
@@ -432,6 +432,51 @@
     %code
       https://{{ domain }}{% url 'handlers.hub9_handler' 'delivered' object.uuid %}
 
+  -elif object.channel_type == 'FCM'
+    %h4
+      -trans "To finish configuring your Firebase Cloud Messaging connection you'll need to send the data to URL with parameters `from`, `msg` and `fcm_token` to receive messages and to register contacts use `urn`, `fcm_token` and `name` (optional)."
```
nicpottier: To use your Firebase Cloud Messaging channel you'll have to POST to the following URLs with the parameters below.

temba/contacts/models.py
```diff
@@ -1762,22 +1776,24 @@ class ContactURN(models.Model):
                                 help_text="The preferred channel for this URN")
 
     @classmethod
-    def get_or_create(cls, org, contact, urn_as_string, channel=None):
+    def get_or_create(cls, org, contact, urn_as_string, channel=None, extra_path=None):
```
nicpottier: Don't understand what's going on with `extra_path`, can you explain this a bit?

temba/channels/models.py
```diff
@@ -323,6 +328,25 @@ def add_viber_channel(cls, org, user, name):
...
+    def add_fcm_channel(cls, org, user, data):
+        """
+        Creates a new Firebase Cloud Messaging channel
+        """
+        from temba.contacts.models import FCM_SCHEME
+
+        assert Channel.CONFIG_FCM_KEY in data and Channel.CONFIG_FCM_TITLE in data, "%s and %s are required" % (
+            Channel.CONFIG_FCM_KEY, Channel.CONFIG_FCM_TITLE)
+
+        existing = Channel.objects.filter(is_active=True, org=org, channel_type=Channel.TYPE_FCM).first()
```
nicpottier: Why force only one FCM channel per account?

temba/channels/models.py
```diff
@@ -323,6 +328,25 @@ def add_viber_channel(cls, org, user, name):
...
+        if existing:
+            existing.config = json.dumps(data)
+            existing.save(update_fields=('config',))
+            return existing
+        else:
+            return Channel.create(org, user, None, Channel.TYPE_FCM, name=org.name, address="fcm-%s" % org.slug,
```
nicpottier: Is there any kind of validation we can do of the tokens they have handed us? Always nice to fail fast when the user is inputting stuff to make sure they put in the right tokens.

temba/channels/models.py
```diff
@@ -1105,6 +1129,48 @@ def success(cls, channel, msg, msg_status, start, request_method=None, request_u
...
+
+        if api_key:
+            url = 'https://fcm.googleapis.com/fcm/send'
+            title = channel.config.get(Channel.CONFIG_FCM_TITLE)
+            data_message = {'type': 'rapidpro', 'title': title, 'message': text, 'message_id': msg.id}
+            payload = json.dumps({
+                'data': data_message,
+                'content_available': True,
+                'to': msg.urn_path,
+                'notification': {
```
nicpottier: So if I understand this right, messages from RapidPro will always display as a notification, is that right?

Should that be an option when adding the channel as to whether you want notifications or pure data?

temba/channels/models.py
```diff
@@ -1105,6 +1129,48 @@ def success(cls, channel, msg, msg_status, start, request_method=None, request_u
...
+                else:
+                    raise SendException("Got non-200 response [%d] from Firebase Cloud Messaging" %
+                                        response.status_code,
+                                        method='POST',
+                                        url=url,
+                                        request=payload,
+                                        response=result,
+                                        response_status=response.status_code)
+            except Exception as e:  # pragma: no cover
+                ChannelLog.log_error(msg, e)
```
nicpottier: Good to log what was actually posted (and what any possible response was) here, aids in debugging.

temba/channels/urls.py
```diff
@@ -46,6 +46,7 @@
         url(r'^/globe/(?P<action>receive)/(?P<uuid>[a-z0-9\-]+)/?$', GlobeHandler.as_view(), name='handlers.globe_handler'),
         url(r'^/viber/(?P<action>status|receive)/(?P<uuid>[a-z0-9\-]+)/?$', ViberHandler.as_view(), name='handlers.viber_handler'),
         url(r'^/line/(?P<uuid>[a-z0-9\-]+)/?$', LineHandler.as_view(), name='handlers.line_handler'),
+        url(r'^/fcm/(?P<action>register|receive)/(?P<uuid>[a-z0-9\-]+)/?$', FCMHandler.as_view(), name='handlers.fcm_handler'),
```
nicpottier: We'll merge upstream today, but this has changed slightly be simpler.

temba/contacts/models.py
teehamaral: The FCM token on mobile device is updated frequently, so, the contact URN should be updated on RapidPRO as well.
The `extra_path` would be used to prevent duplicated contact, updating only path field instead create a new URN or a new contact. This was our first solution. What would be the best way to resolve this problem?

temba/channels/models.py
teehamaral: Unfortunately, we didn't find any API to validate the token =(

teehamaral:
Review, please.

nicpottier:
Ok so I think I understand what's going on here.

We shouldn't change the meaning of `path` or `scheme` on URN, that's part of what URNs are, so they should always mean the identity of the URN on the path side.

As I understand it the firebase ids are essentially `[id]:[auth_token]` where `id` doesn't change right? So that should always be the path for these URNs. I think we should just add a nullable `auth` field to ContactURN to store the `auth_token` portion. We'll need to pass that down to the send message task etc, when needed. But otherwise the rest should be the same as you have it.

But ya, you don't want to change the meaning or encoding of what a URN is, which should be about identity, not authentication.

nicpottier:
@rowanseymour other thoughts here?

teehamaral:
Review, please.

temba/contacts/models.py
```diff
@@ -1611,6 +1621,14 @@ def get_urn(self, schemes=None):
             # otherwise return highest priority of any scheme
             return urns[0] if urns else None
 
+    def update_urn_path(self, urn_as_string, path):
```
rowanseymour: This doesn't seem to be used any more?

temba/channels/tests.py
```diff
@@ -8505,7 +8531,8 @@ def test_send(self):
             Channel.send_message(dict_to_struct('MsgStruct', msg.as_task_json()))
 
             mock.assert_called_once_with('https://devapi.globelabs.com.ph/smsmessaging/v1/outbound/21586380/requests',
-                                         headers={'User-agent': 'RapidPro'},
+                                         headers={'User-agent': 'RapidPro', 'Content-Type': 'application/json',
+                                                  'Authorization': u'key=123456789'},
```
rowanseymour: I think these extra headers are coming from a test leaking changes to TEMBA_HEADERS

temba/channels/models.py
```diff
@@ -1188,6 +1207,62 @@ def success(cls, channel, msg, msg_status, start, request_method=None, request_u
...
+        if result and 'success' in result and result.get('success') == 1:
+            external_id = result.get('multicast_id')
+            Channel.success(channel, msg, WIRED, start, 'POST', url, payload, response, external_id)
+        else:
+            raise SendException("Got non-200 response [%d] from Firebase Cloud Messaging" %
+                                response.status_code,
+                                method='POST',
+                                url=url,
+                                request=payload,
+                                response=result,
```
nicpottier: response here should probably be `response.content` in case the status code isn't 200 so we can get the body regardless

temba/contacts/models.py
teehamaral: Oh, you are right. It was another solution, sorry. I'll remove it.

teehamaral:
Review, please.

temba/msgs/models.py
```diff
@@ -1151,7 +1151,7 @@ def as_task_json(self):
                     status=self.status, direction=self.direction,
                     external_id=self.external_id, response_to_id=self.response_to_id,
                     sent_on=self.sent_on, queued_on=self.queued_on,
-                    created_on=self.created_on, modified_on=self.modified_on)
+                    created_on=self.created_on, modified_on=self.modified_on, auth_token=self.contact_urn.auth_token)
```
teehamaral: Adding `auth` in `dict` just when is it set? As like:
```
task_json = dict(id=self.id, org=self.org_id, channel=self.channel_id, broadcast=self.broadcast_id,
                 text=self.text, urn_path=self.contact_urn.path,
                 contact=self.contact_id, contact_urn=self.contact_urn_id,
                 priority=self.priority, error_count=self.error_count, next_attempt=self.next_attempt,
                 status=self.status, direction=self.direction,
                 external_id=self.external_id, response_to_id=self.response_to_id,
                 sent_on=self.sent_on, queued_on=self.queued_on,
                 created_on=self.created_on, modified_on=self.modified_on)

if self.contact_urn.auth:
    task_json.update(dict(auth=self.contact_urn.auth))

return task_json
```

templates/channels/channel_configuration.haml
```diff
@@ -510,6 +510,47 @@
...
+        .value {{ object.name }}
+      .config-value
+        .name
+          -trans "Address/Key:"
+        .value {{ object.address }}
+      .config-value
+        .name
+          -trans "Notification title:"
+        .value {{ object.config_json.FCM_TITLE }}
+
```
teehamaral: The channel settings, or are they unnecessary here?

temba/msgs/models.py
nicpottier: Yep

templates/channels/channel_configuration.haml
nicpottier: Well what are these used for? If there isn't any configuration the user has to do beyond hooking up the channel then we shouldn't be showing them things. They need to set a callback on the Firebase side though ya?

teehamaral: Yes, it's need set the key in the app. But, it's already displayed as channel address, what do you think?

nicpottier: Man, I'm sorry, I'm just not getting this for some reason.

So if someone connects a Firebase channel, from their Android app, what URLs will they be hitting? Do they actually hit the "contact register" and "receive urls" themselves or does Firebase do that?

My impression is they are hitting Firebase and Firebase hits us, is that not the case? Sorry, the firebase docs aren't super clear on this call flow either..

teehamaral: The Android app, for example, will call directly to RapidPRO (register and receive), Firebase will be used as a way of communication between RapidPRO and the app, precisely because of its notification feature. 

On the Firebase side, we just generate the key. So, we add a new FCM channel in RapidPRO, which will generate the URLs to integrate with the app, the app will listen to Firebase to receive the messages from RapidPRO throught user identifier in the app and use the URLs to send messages and register contacts on RapidPRO.

templates/channels/channel_configuration.haml
```diff
@@ -510,6 +510,33 @@
...
+      -trans "Receive URL"
+
+    %p
+      -trans "To handle incoming messages, POST to the following URL with the parameters from, msg and fcm_token."
+
+    %code
+      https://{{ domain }}{% url 'handlers.fcm_handler' 'receive' object.uuid %}
+
+    %hr
+
```
nicpottier: Ah, ok, after doing some more reading I understand now that we aren't adding upstream messaging here (device to RapidPro via Firebase) only downstream. MO messages are delivered directly via the HTTP handler above and hit directly by the Android app, is that right?

teehamaral: Right

nicpottier:
Thanks guys!

teehamaral:
You're welcome.

PR #410, opened 2016-12-19
Multichannel triggers

nullcode:
Let's continue the discussion about multichannel triggers here @ericnewcomer @nicpottier 

With USSD triggers I don't see the point to have the same trigger with more channels at the same time, since keywords/starcode will be unique per channel. However you guys mentioned something that other trigger types would have good use for it.

Generally what I have PRed now would be enough for USSD cases as far as I know.


nullcode:
@nicpottier @ericnewcomer can you take a look at this?

nicpottier:
I think this is correct, I'm going to change this in our branch with some other Trigger work I'm doing and merge it up though since it is behind where we are at.

PR #435, opened 2017-01-30
Junebug channel for RapidPro.

smn:
I've taken the Jasmin channel as an example and have modelled the Junebug channel along the same lines. 

As Junebug can support both SMS and USSD in the same channel type I suspect more work is required here to get this to work as a USSD channel in RapidPro.

This should be enough to get it to work with SMS.


templates/channels/channel_claim.haml
```diff
@@ -423,6 +423,22 @@
...
+            .btn-container
+              .claim-glyph.icon-channel-external
+              %p.claim-text
+                -trans "Junebug"
+
+        .span9
+          .claim-description
+            %div
+              -blocktrans
+                Connect your <a href="http://junebug.praekelt.org/" target="_blank">Junebug</a> instance that you have
```
bruskiza: Should we not be prompting https here just for good measure?

temba/channels/handlers.py
```diff
@@ -1730,6 +1730,93 @@ def post(self, request, *args, **kwargs):
...
+                        ', '.join(expected_keys)), status=400)
+
+            content = data['content']
+            message = Msg.create_incoming(
+                channel, URN.from_tel(data['from']), content)
+            Msg.objects.filter(pk=message.id).update(
+                external_id=data['message_id'])
+            return HttpResponse('OK')
+
+        else:  # pragma: needs cover
```
rowanseymour: - 'needs cover' is for existing code that we haven't yet written tests for so that we can keep coverage at '100%' - makes it easier to see where coverage has changed on a PR.
- 'no cover' is for stuff those few odd things we're ok with not testing

.travis.yml
```diff
@@ -15,6 +15,7 @@ install:
 - pip install --upgrade pip
 - pip --version
 - pip install -r pip-freeze.txt --upgrade
+- pip install setuptools==33.1.1  # Recent version breaks packaging.requirements, this is a temp fix: https://github.com/pypa/setuptools/issues/942
```
rowanseymour: Have you seen this fail on Travis? https://github.com/pypa/setuptools/issues/942#issuecomment-274817619 Seems to suggest we should be ok with latest pip here.

smn: it was failing last week when I wrote this, I'll remove it again as hopefully it's been fixed now. Thanks for spotting.

temba/channels/handlers.py
smn: Thanks, I'll remove these and attend to the coverage. I suspect this is due to me following the Jasmin example.

templates/channels/channel_claim.haml
smn: done

temba/channels/handlers.py
smn: I've fixed these with the last commit, this last one I've removed entirely as invalid actions are currently guarded against by the urlconf pattern for the handler.

.travis.yml
smn: Still failing on Travis, I've put this back for now.

smn:
@rowanseymour thanks for the initial review, should be ready for another pass.

smn:
Thanks! 🍰  

Will you merge and if not, please let me know what else is expected from me :)

nicpottier:
We will need a signed CAA from you before merging.

smn:
Cool, working on this internally.

smn:
Submitted via email

temba/channels/migrations/0050_add_junebug_channel_type.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
nicpottier: Just merged from us because the migration squashing wasn't on /rapidpro yet and redoing that wasn't in the cards. Shouldn't affect this PR except that this migration should be 0058 and depend on 0057

smn:
I'll look into the coverage drop.

smn:
@nicpottier changes applied, ready for re-review.

nicpottier:
Great, will merge this today. Does Junebug have a squarish logo we can use on the channel claim page? If so I'll add this to the font icon set so it can be all pretty-like.

smn:
We've got https://github.com/praekelt/junebug-marketing/blob/master/html/img/junebug-logo.svg but that's not quite square. I'm checking to see if we have something square.

nicpottier:
Nice ya that would work if maybe we just had the Jun squiggle piece?

nicpottier:
Merged this just to get it in (since we are moving super quick right now) but would still love the squiggle and will work that in if you get it to me!

PR #445, opened 2017-02-02
Use /sitestatic/ as the STATIC_URL in settings_common.py

smn:
[settings_common.py](https://github.com/rapidpro/rapidpro/blob/844605262179768aad6344bb770a368128380da9/temba/settings_common.py#L120) sets `/static/` as the `STATIC_URL` which [settings.py.dev](https://github.com/rapidpro/rapidpro/blob/844605262179768aad6344bb770a368128380da9/temba/settings.py.dev#L88) changes to `/sitestatic/`

This results in [`COMPRESS_OFFLINE_CONTEXT` to reference](https://github.com/rapidpro/rapidpro/blob/844605262179768aad6344bb770a368128380da9/temba/settings_common.py#L1155) the old `/static/` instead of `/sitestatic` which causes problems during offline compression.


nicpottier:
Hrmm.. I'm a bit confused as to what's happening here. Are you compressing with one setting and running with another? How do these get out of wack?

Maybe walking me through what the repro steps are would help..

smn:
I came across this while trying to address #446. If you're using the `settings.py.dev` rapidpro ships with as the settings file your `COMPRESS_OFFLINE_CONTEXT` will reference the wrong `STATIC_URL` since it'll use the one defined in `settings_common` instead of what's in `settings.py` because of how those settings are ordered on import.

```
$ ./manage.py shell -i python
Python 2.7.12 (default, Oct 11 2016, 05:20:59)
[GCC 4.2.1 Compatible Apple LLVM 8.0.0 (clang-800.0.38)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
(InteractiveConsole)
>>> from temba import settings
>>> settings.STATIC_URL
u'/sitestatic/'
>>> settings.COMPRESS_OFFLINE_CONTEXT
{'base_template': u'frame.html', 'STATIC_URL': u'/static/'}
>>>
```

For compressor offline to work properly it needs to have `/sitestatic/` as its `STATIC_URL`

nicpottier:
So we redid a few things to make sure compression was doing the right thing and part of that was discovering this same problem so going to close this as I believe it is addressed. Thanks for bringing it to our attention.

smn:
Yeah fixed, thanks.

PR #458, opened 2017-02-08
Enable searching in groups and blocked/stopped contacts

rowanseymour:
Currently the search box is shown at the top of any contact listing page, but using it always takes the user back to the 'All Contacts' list page and does a search on that group.


rowanseymour:
Ugh wrong repo

Issue #453, opened 2017-02-07
Size characters limit alert 


nelsonlapf:
Hi guys,

Yesterday I got a feedback from France saying that would be good if we could implement a way to alert them when they are exceeding the numbers of characters during a flow construction, for example, the field Category that only allows them to use 35 characters, would be good if we could implement a counter, a red alert or just don´t allow them to write more.

What do you think?

nicpottier:
Doesn't the message dialog already show the number of characters and messages?

nelsonlapf:
I mean here:

![capturar](https://cloud.githubusercontent.com/assets/23530401/22695362/af89220c-ed31-11e6-8121-a3dfbc11e5e9.JPG)

nicpottier:
I think the limit on the Ruleset label is 64 chars.. which I think we do enforce in the UI. (you can't enter longer than that)

Agree the labels should probably have a sane limit as well, perhaps 36 chars.

nelsonlapf:
That´s the point, for "categorize as" the limit is 35 characters, however, it didn´t stop the user while he is typing and allows him to type as much as he wants, and when it goes to U-Report, it didn´t go right.

So would be good to do the same as for "Save result as" and don´t allow the user to type more, or have a kind of alert.

PR #461, opened 2017-02-13
Lenght on "categorize as"

teehamaral:
This PR includes the limit on "categorize as" to 35 characters (#453).


nicpottier:
@ericnewcomer plz review

teehamaral:
Sometimes the users write more than 35 characters, so they think they're saving everything, but only 35 are saved. Adding this limit, we can avoid this problem.

ericnewcomer:
Ha, Nic found where this value is indeed truncated to 36 characters when saving the flows. So it is indeed a RapidPro inconsistency. That said, these should be bumped to the real limit of 36.

nicpottier:
Sweet, thanks guys!

teehamaral:
🙂

Issue #429, opened 2017-01-18
USSD Flows: Ability to 'Start someone else in a flow'


RichardHeslopZA:
We have a common requirement on our projects for a healthcare worker to capture a patient as a new contact over USSD. 

The Documentation (http://docs.rapidpro.io/#article_366849) describes how to achieve this on a Messaging flow (using the 'Start Someone in a Flow' action). However, this functionality is not available in 'USSD Messaging' flows. At present, it does not seem possible to register another Contract from within a USSD flow.

SeanBlaschke:
Wanted to bump this and see if feature is expected to be included in any upcoming releases. Thanks!

ewheeler:
@SeanBlaschke related mailing list discussion:  https://groups.google.com/forum/#!topic/rapidpro-dev/fBHwVplLpGM

smn:
Pull request ready for review in #464 @nicpottier & @ewheeler

nicpottier:
Believe this should be fixed now.

PR #464, opened 2017-02-21
allow USSD flows to start someone else in a flow

smn:
rapidpro_-_visually_build_interactive_sms_applications

I've not tested this on a phone yet but this does expose the required functionality in the UI.

This addresses issue #429


smn:
tests are failing due to migrations unrelated to this PR.

nicpottier:
We'll have an upmerge to fix tests but will need a test for this functionality before we merge as well.

smn:
@nicpottier happy to add a test. Could you provide some guidance at what level would you like this to be tested in a way that it currently isn't yet? The model level test I'm seeing relates to flows migrating to version 9 of the flow spec, other than that I'm not immediately seeing a test that I can use as a starting point.

nicpottier:
Any of the FlowFileTests in /temba/flows/tests.py should be good examples. Test should start a USSD flow that starts a Messaging flow as an action.

jpaidoussi:
@nicpottier @smn It looks like something I did made this PR's diff unreadable even though it should be a small change compared to the master branch. If you close this PR I will create a new one from our branch which should have the right diff.

smn:
Yeah also confused, closing this PR. Please reference this one in the one you're creating?

Issue #469, opened 2017-02-24
Vumi Channel does not send the `in_reply_to` value for USSD sessions.


smn:
Depending on the interface to the aggregator the value of the `in_reply_to` is important as for both synchronous and asynchronous protocols it is often a requirement to be able to refer to a previous request.

The value of the `in_reply_to` value should be the `message_id` of the previous message that was received to which this is a reply.

nicpottier:
Should be a pretty small change in send_vumi_message in channels/models.py.. We should be tracking the message the response is to so that id is available on the msg dict.

smn:
https://github.com/rapidpro/rapidpro/pull/472 addresses this issue.

PR #472, opened 2017-02-24
Use the reply_to value from the message as in_reply_to

smn:
Depending on the interface to the aggregator the value of the `in_reply_to` is important as for both synchronous and asynchronous protocols it is often a requirement to be able to refer to a previous request.

The value of the `in_reply_to` value should be the `message_id` of the previous message that was received to which this is a reply.

Addresses issue #469


temba/channels/models.py
```diff
@@ -2014,7 +2014,8 @@ def send_vumi_message(cls, channel, msg, text):
         channel.config['transport_name'] = 'ussd_transport' if is_ussd else 'mtech_ng_smpp_transport'
 
         payload = dict(message_id=msg.id,
-                       in_reply_to=None,
+                       in_reply_to=(msg.response_to.external_id
+                                    if msg.response_to_id else None),
```
smn: Note that I'm checking `response_to_id` here instead of `response_to` as the `MsgStruct` created only knows about `response_to_id` and it's not a Django Model so it doesn't expose the value like a ForeignKey would. This is also why I'm creating a nested MsgStruct in the test with both `response_to` and `response_to_id` keys.

smn:
Ready for review

temba/channels/models.py
nicpottier: Huh, I must be confused. How do you expect this to work in real life when MsgStruct doesn't have the value you are stuffing in it in the test?

At this point the Msg is just a struct since we don't hit the db when sending messages (except to update status).. so we either need to do a real lookup for this external id from the Msg model (which hopefully we would only do over USSD) or stuff the external id at the point we create the msg struct. (some careful select_related might not make this too painful)

smn: Right, this now makes more sense to me. I was thinking the MsgStruct was a thing only used for testing to mock out a Message. I'll update.

smn:
@nicpottier please have a look again, I'm looking up the `response_to` msg if the `response_to_id` is specified in the struct. Not sure if you're happy with the db lookup at this stage but it seemed less invasive than changing the struct you're working with for message sending.

temba/channels/models.py
```diff
@@ -2007,14 +2007,19 @@ def send_smscentral_message(cls, channel, msg, text):
...
     def send_vumi_message(cls, channel, msg, text):
-        from temba.msgs.models import WIRED
+        from temba.msgs.models import WIRED, Msg
         from temba.contacts.models import Contact
 
         is_ussd = channel.channel_type in Channel.USSD_CHANNELS
         channel.config['transport_name'] = 'ussd_transport' if is_ussd else 'mtech_ng_smpp_transport'
 
+        if msg.response_to_id:
+            in_reply_to = Msg.objects.get(pk=msg.response_to_id).external_id
```
nicpottier: Since this is a tight loop, let's do:

`in_reply_to = Msg.objects.values_list('external_id', flat=True).get(pk=msg.response_to_id)`

nicpottier:
Nice, fix that build and we're good to go.

smn:
yeah, failing to reproduce these errors locally. Looking into it.

temba/channels/models.py
```diff
@@ -2013,10 +2013,8 @@ def send_vumi_message(cls, channel, msg, text):
         is_ussd = channel.channel_type in Channel.USSD_CHANNELS
         channel.config['transport_name'] = 'ussd_transport' if is_ussd else 'mtech_ng_smpp_transport'
 
-        if msg.response_to_id:
-            in_reply_to = Msg.objects.get(pk=msg.response_to_id).external_id
-        else:
-            in_reply_to = None
+        in_reply_to = Msg.objects.values_list(
+            'external_id', flat=True).get(pk=msg.response_to_id)
```
nullcode: @nicpottier @smn I just stumbled upon this and I don't think that's going to work, because values_list will return a QS with external_ids only, try Msg.objects.only('external_id') that will select only the external_id and pk columns. Also the .get() might throw a big exception here.The tests might not be complete enough if they haven't popped this up, unless I'm mistaken. I'm doing something very similar right now for the session and I moved this to a separate method so it's easier to test, looks like this:
```   
def get_session_with_status_only(self, session_id):         
        return self.only('status').filter(id=session_id).first()

Comment (nicpottier) [inline temba/channels/models.py]: @nullcode this is actually straight out of the Django docs and the recommended way of getting a single value from a FK:
  https://docs.djangoproject.com/en/1.10/ref/models/querysets/#values-list

Since it is a foreign key and we never delete incoming messages the lookup should never fail from a GET perspective.

Comment (nullcode) [inline temba/channels/models.py]: @nicpottier ah my bad then, I must have used it wrong when I tried then. I wonder if it has the same performance as only or defer. Sorry if I created confusion, I just felt the need to comment, because I'm working on something very similar.

Comment (nicpottier) [inline temba/channels/models.py]: Also was somewhat dubious of what kind of SQL this would turn into, but just tested it and it is indeed right. Django is weird that way.

```python
Msg.objects.values_list('external_id', flat=True).get(pk=42)
```

```sql
SELECT "msgs_msg"."external_id" FROM "msgs_msg" WHERE "msgs_msg"."id" = 42
```

nicpottier: @nullcode Using 'only', at least as below turns into two queries which is kinda weird:

```python
Msg.objects.filter(id=42).only('external_id').first()
```

```sql
SELECT "msgs_msg"."id", "msgs_msg"."external_id" FROM "msgs_msg" WHERE "msgs_msg"."id" = 42 ORDER BY "msgs_msg"."created_on" DESC, "msgs_msg"."id" DESC LIMIT 1
SELECT "msgs_msg"."id", "msgs_msg"."text" FROM "msgs_msg" WHERE "msgs_msg"."id" = 42
```

But you are paying the overhead of Django creating the object, which isn't insignificant in tight loops.

nicpottier: Hah, oh nevermind, that second query is caused by the print in my console.

So ya, the only is pretty similar, though a bit heavier than the `values_list` appraoch.

nullcode: wow thanks for following this up, it is very interesting. I need the object too so I can update the field I selected so its beneficial in my case. The values_list approach is a bit counterintuitive for me, especially cos' values_list is actually a lazy QS that returns a list of tuples or single values when flattened but you can still use get/filter on anything you haven't selected in the list.

smn:
@nicpottier build fixed.

nicpottier:
No looks good, we'll merge this and the other PR today, just need Travis to start cooperating. (we have a bunch of things upstream we'll merge in first but I'll take care of conflits)

smn:
Thanks!

PR #475, opened 2017-03-01
Feature/issue 429 start someone in a flow

jpaidoussi:
This is a replacement for https://github.com/rapidpro/rapidpro/pull/464 due to the broken diff there. @nicpottier I have added a test that I think should cover the case you wanted but it doesn't appear to work. Unfortunately I don't have the context to tell if is something wrong with the test or the underlying code.

If you have a minute to look over it and let me know that would be much appreciated.


temba/ussd/tests.py
```diff
@@ -443,3 +444,40 @@ def test_interrupt(self, create_incoming):
...
+    def test_ussd_subflow(self):
+        # start a session
+        callback_url = reverse('handlers.vumi_handler', args=['receive', self.channel.uuid])
+        ussd_code = "*111#"
+        data = dict(timestamp="2016-04-18 03:54:20.570618", message_id="123456", from_addr="+250788383383",
+                    content="None", transport_type='ussd', session_event="new", to_addr=ussd_code)
+        response = self.client.post(callback_url, json.dumps(data), content_type="application/json")
+        child_flow = self.get_flow('ussd_subflow_child')
+        flow = self.get_flow('ussd_subflow_parent', dict(NEW_CONTACT_FLOW_ID=child_flow.uuid))
+
```
nicpottier: If I understand this correctly, should we be validating here that no FlowRun was created from the incoming message since no Trigger was configured?

jpaidoussi: @nicpottier Like this https://github.com/rapidpro/rapidpro/pull/475/commits/6ba50083a6731a7b9559f6220c3b0cf3a6f07383 ?

temba/ussd/tests.py
```diff
@@ -443,3 +444,41 @@ def test_interrupt(self, create_incoming):
...
+    def test_ussd_subflow(self):
+        # start a session
+        callback_url = reverse('handlers.vumi_handler', args=['receive', self.channel.uuid])
+        ussd_code = "*111#"
+        data = dict(timestamp="2016-04-18 03:54:20.570618", message_id="123456", from_addr="+250788383383",
+                    content="None", transport_type='ussd', session_event="new", to_addr=ussd_code)
+        response = self.client.post(callback_url, json.dumps(data), content_type="application/json")
+        child_flow = self.get_flow('ussd_subflow_child')
+        flow = self.get_flow('ussd_subflow_parent', dict(NEW_CONTACT_FLOW_ID=child_flow.uuid))
```
ericnewcomer: If these two flows are in the same import file, then this sort of mapping is unnecessary. They'll get new uuids, relatively replaced during the flow import itself.

smn:
Looks like the Travis & Coveralls issues yesterday cause this build's coveralls reports to get stuck somewhere. @nicpottier or @ericnewcomer could either of you restart this build on Travis to get the missing coveralls check to complete?

smn:
🎉  coveralls check seems sorted now

Issue #430, opened 2017-01-18
USSD Flows: Channel selection bugs


RichardHeslopZA:
Environment: RapidPro.io

We may have hit a few more USSD teething issues in terms of an account configured with both a USSD and SMS channel. On our account, we have added both the Vumi USSD channel and a Nexmo account (South African number - +27).

We have noticed:
1) In the Contact detail view, the 'Send Message' button causes the system to attempt to send the message over the USSD channel. We can see the attempted message send over USSD in the Message History.
2) If we build a Messaging flow and use a timer Trigger to start the flow, the system also attempts to send the messages over USSD. We can see the attempted message send (Broadcast) over USSD in the Message History.

SeanBlaschke:
Has there been any progress adressing this bug? We are set to launch a national programme in 2 weeks with the Ministry of Health and this currently is one of 3 showstoppers.

nicpottier:
Should be fixed now. You may need to remove/readd existing USSD channels for it to take effect.

PR #423, opened 2017-01-06
USSD session - graceful termination

nullcode:
I wrote a DFS based algorithm that checks whether a session should be closed at a certain point in a flow. The question is how should I pass that information to the SEND_FUNCTIONS. 

Since I don't want to mess around with new fields on Msg I leave it as an open question for you guys @ericnewcomer @nicpottier. My idea is to pass it as an extra property of the runtime created `MsgStruct` that is pushed to `task_msgs`, but you guys might know better.

PR #477, opened 2017-03-02
Junebug USSD channel

smn:
This adds a second Junebug channel type, this one is for USSD.


temba/channels/models.py
```diff
@@ -1322,28 +1325,44 @@ def send_jasmin_message(cls, channel, msg, text):
...
+
         # build our payload
         payload = dict()
-        payload['to'] = msg.urn_path
         payload['from'] = channel.address
         payload['event_url'] = event_url
         payload['content'] = text
 
+        if is_ussd:
+            external_id, session_status = Msg.objects.values_list(
```
nicpottier: Don't wrap at 80.. we go to 120 around here. :)

smn:
@nicpottier thanks, that was my next question. Junebug supports a bunch of modalities that extend beyond a single URN type. I'll look at the Twilio handler to see how that's done there.

temba/channels/models.py
smn: 👍

smn:
I'm not quite sure what's going on with the migration that was created, it touches more fields than I've changed in this PR. The only field change I would have expected was the addition to the `channel_type`.

smn:
@nicpottier after merging in the latest changes I get `makemigrations` suggesting a lot of changes to `created_on` and `modified_on` fields and these are failing the build. Here's the full list: https://gist.github.com/smn/38659d2bcd8dd5510c0e81a557ec56f5. I suspect the channels app is missing from the list because this migration's added those.

Any idea where these are coming from?

smn:
Ah, found the smartmin upgrades in v3.0.66.

smn:
@nicpottier ready for review

temba/channels/models.py
```diff
@@ -1310,30 +1313,43 @@ def send_jasmin_message(cls, channel, msg, text):
...
-        from temba.msgs.models import WIRED
+        from temba.msgs.models import WIRED, Msg
 
         # the event url Junebug will relay events to
         event_url = 'https://%s%s' % (
             settings.HOSTNAME,
             reverse('handlers.junebug_handler',
                     args=['event', channel.uuid]))
 
+        is_ussd = channel.channel_type in Channel.USSD_CHANNELS
```
nicpottier: No reason to iterate the list on every msg send if we can just test against the channel type.

temba/channels/handlers.py
```diff
@@ -1991,12 +2004,45 @@ def post(self, request, *args, **kwargs):
                     "Missing one of %s in request parameters." % (
                         ', '.join(expected_keys)), status=400)
 
-            content = data['content']
-            message = Msg.create_incoming(
-                channel, URN.from_tel(data['from']), content)
-            Msg.objects.filter(pk=message.id).update(
-                external_id=data['message_id'])
-            return HttpResponse('OK')
+            if is_ussd:
```
nicpottier: Can we de-80 char this? :)

smn: doh. sorry, editor defaults & old habits die hard.

smn:
bunch of USSDSession tests now breaking, working on fixing those. Will ping you when ready for a review again.

temba/channels/views.py
```diff
@@ -1702,9 +1706,30 @@ class JunebugForm(forms.Form):
...
+
+            if not org:  # pragma: no cover
+                raise Exception("No org for this user, cannot claim")
+
+            data = form.cleaned_data
+
+            role = {
+                Channel.TYPE_JUNEBUG: Channel.ROLE_SEND + Channel.ROLE_RECEIVE,
+                Channel.TYPE_JUNEBUG_USSD: Channel.ROLE_USSD,
+            }.get(data['channel_type'])
```
nicpottier: clever but I'd prefer a simple `if`?

temba/channels/views.py
```diff
@@ -1702,9 +1706,30 @@ class JunebugForm(forms.Form):
...
         title = _("Connect Junebug")
-        channel_type = Channel.TYPE_JUNEBUG
         form_class = JunebugForm
-        fields = ('country', 'number', 'url', 'username', 'password')
+        fields = ('channel_type', 'country', 'number', 'url', 'username', 'password')
+
+        def form_valid(self, form):
+            org = self.request.user.get_org()
+
+            if not org:  # pragma: no cover
```
nicpottier: This is garbage (that needs to be removed elsewhere as well), you can leave it out.

temba/channels/tests.py
```diff
@@ -2979,6 +2979,62 @@ def test_claim_chikka(self):
...
+            "number": "+273454325324",
+            "url": "http://example.com/messages.json",
+            "username": "foo",
+            "password": "bar",
+        }
+
+        response = self.client.post(reverse('channels.channel_claim_junebug'), post_data)
+
+        channel = Channel.objects.get()
+        self.assertEquals(channel.channel_type, Channel.TYPE_JUNEBUG_USSD)
```
nicpottier: Maybe add an assert on the role here?

temba/channels/models.py
```diff
@@ -137,6 +138,9 @@ class Channel(TembaModel):
     ROLE_RECEIVE = 'R'
     ROLE_CALL = 'C'
     ROLE_ANSWER = 'A'
+    ROLE_USSD = 'U'
+
+    DEFAULT_ROLE = ROLE_SEND + ROLE_RECEIVE
```
nicpottier: 👍

temba/channels/tests.py
smn: done

temba/channels/views.py
smn: done

temba/channels/views.py
smn: done

temba/channels/models.py
```diff
@@ -1310,30 +1317,43 @@ def send_jasmin_message(cls, channel, msg, text):
...
         payload['from'] = channel.address
         payload['event_url'] = event_url
         payload['content'] = text
 
+        if is_ussd:
+            external_id, session_status = Msg.objects.values_list('response_to__external_id', 'session__status').get(pk=msg.id)
+            # NOTE: Only one of `to` or `reply_to` may be specified
+            payload['reply_to'] = external_id
+            payload['channel_data'] = {
+                'continue_session': session_status not in ChannelSession.DONE,
```
smn: @nullcode I noticed that in #423 you've added a boolean property to session called `should_end` that checks for the `status` being in state `ENDING`. Is this perhaps what I should be using here too? Is `COMPLETED` as a status only set after the message has been sent and am I checking that too early at this point?

smn:
@nicpottier ready for review again, would like some input from @nullcode (or other people working on this) on how to recognise a session that's ending.

temba/channels/models.py
nullcode: @smn yes, use that please. I just pushed the last part of the PR, you can check how I did it with Vumi's send_message, it only marks it completed when the aggregator responds with success.

smn:
@nicpottier I'll be adding the extra contract tests in flows/tests.py today. How do you suggest we proceed on the session closing flag stuff? Ideally I'd like to build on the work that @nullcode is doing in #423 but not sure how soon that'll merge in. Other option is to merge this without support for gracefully ending a session which is meh.

smn:
Two tests added that check that `get_send_channel()` and `get_ussd_channel()` return channels with the correct roles.

nicpottier:
I'm cool merging this in after we upmerge Nyaruka (in a few mins). I think we can get @nullcode to make the session ending changes and maybe add in the USSD role to the Vumi channel while he's at it which would then bring everything up to parity.

smn:
@nicpottier fixed the migrations but it's failing on an unrelated Twilio test now.

nicpottier:
There's some weird race condition / binding issue with some mocking on that test. I'll look at figuring out how to fix it.

Since you are adding the new USSD role, can you have it set when the USSD Vumi channels is created:
  https://github.com/rapidpro/rapidpro/blob/master/temba/channels/views.py#L2004

That'll address #430

nicpottier:
Looks like there's a vumi failure now, which I don't think is the same as the random Twilio test failure..

smn:
@nicpottier yeah, the Twilio one's gone now.

smn:
@nicpottier the test is fixed now but the way the current code layout allows for it is a bit ugly. Ideally I'd like to set the message types as attributes on the `Msg` model and skip the need for importing these separate in various places. I was thinking of doing that refactor in a different PR as i'm not sure of how much of the code base that will touch and it'll potentially be a bit more difficult to review as part of this PRs work.

Thoughts?

smn:
See new failures now on the last commit. Will attend to those. I've got the refactor on the Msg attributes ready in a different branch which I'll PR shortly.

nicpottier:
Sounds good, what exactly are you thinking about with the Msg refactor?

PR #485, opened 2017-03-15
Fix Junebug USSD channel tests

smn:
These were originally to be part of #477 but that got merged while I was travelling.


nicpottier:
I fixed this last night on our branch and will upmerge that in a sec.

PR #488, opened 2017-03-16
API v2: Remove all parts of API v1 not used by Surveyor

rowanseymour:
To borrow that famous misquote: "the reports of my demise are greatly exaggerated"

Surveyor uses a few endpoints from API v1 (see [TembaAPI.java](https://github.com/rapidpro/surveyor/blob/master/app/src/main/java/io/rapidpro/surveyor/net/TembaAPI.java)) so am assuming we keep these around for now. I've ripped out everything that isn't used by Surveyor and restricted access to what is left to only surveyor users.


rowanseymour:
Wrong repo argh

PR #491, opened 2017-03-20
Upgrade to Alpine linux friendly version of python-magic

smn:
pip-freeze.txt wants 4.0.12 which breaks on Alpine Linux, 4.0.13 fixes that. 
Alpine is used for the rapidpro docker images and so the docker builds are broken because of the 4.0.12 dependency.


smn:
@nicpottier ready for review, it's just a minor version bump.

PR #492, opened 2017-03-20
Pending messages query

teehamaral:
This PR includes an improvement on the method that check the pending messages from channel.

The celery runs a task each 5 minutes to check the still pending messages, called `check_messages_task`. This function return all messages with status equals `PENDING` that still wasn't sent, but we consider that must exists a minimum time to queued again the messages with this status. So, as solution for this possible issue, was added to query pending messages, the filter in messages creation date.


nicpottier:
Messages get flipped to a QUEUED state once they have been queued, so this shouldn't be catching any messages that are already queued.

Can you explain what this is fixing exactly?

teehamaral:
When a flow is started for 5 thousand contacts, for example, the messages that still being queued have status equals PENDING, if `check_messages_task` is runned, these messages with PENDING status are queued again, in this case, disordered.

nicpottier:
Ah, I see, so it is a race between when the messages are set to pending in `start_msg_flow_batch` and the task being run.

Was thinking another solution might be to not actually update the status of messages to `PENDING` manually in `start_msg_flow_batch` as that would make sure that even with a larger delay we wouldn't get into trouble but I think having the backup of `get_pending_messages` after five minutes makes sense.

Thanks for the contribution!

PR #497, opened 2017-03-27
Show Junebug config for USSD channel and use Radio buttons for selecting channel type.

smn:
This makes use of an `add_config_fields` hook which doesn't seem to be used anywhere else.
Also it addresses (what seems to be) a bug in the `pre_save()` function where every field value was turned into a bool.


nicpottier:
So actually we view channels as immutable on purpose, so we don't want settings to change after the fact. Keeps it very clear when looking at logs that something didn't change out from under us.

smn:
@nicpottier I've removed the channel modification bits, right now it only changes the UI widget for selecting the channel type and fixes a bug where the config for a Junebug USSD channel wasn't displayed.

smn:
thanks!

Issue #499, opened 2017-03-28
Channels listing read links result in 404


smn:
![rapidpro](https://cloud.githubusercontent.com/assets/1065/24418168/62040e3a-13eb-11e7-8d2b-6a740a1bf6b4.png)

The channels have URL patterns setup to use their UUIDs but the template renders the links with `read/<pk>` which results in a page not found.

I suspect this may be something lower down in smartmin templates but I'm not sure which bits of the template to override to fix this.

nicpottier:
Huh, ya, this isn't really a page we ever use. How did you get here?

smn:
![rapidpro](https://cloud.githubusercontent.com/assets/1065/24418997/1666ff8e-13ee-11e7-8be4-75453729782a.png)

The channels icon takes me there, I've got two RapidPro instances running, one is showing the Channels icon and the other isn't. I'm not sure why this is getting triggered on this specific install.

nicpottier:
The way it should work is that you don't see the channels tab unless you don't have a channel, and when you click on it it takes you to the claim page.

smn:
Looking at the [`nav` include](https://github.com/praekeltfoundation/rapidpro/blob/07461f5504742689224362d38bd96392de21463c/templates/includes/nav.haml#L48) Perhaps this is due to me have 3 USSD channels but none of them having the sending role?

rowanseymour:
Isn't this when you're logged in as superuser?

nicpottier:
@smn that definitely sounds likely.

smn:
currently it's filtering on `has_outgoing_channel` should I add USSD channels to that and consider USSD as an outgoing channel or should I just refactor this down to a simple check on whether _any_ channel exists? I'm not sure what the thinking behind this was to limit it to outgoing channels only, I suspect I lack the context on that decision.

smn:
Not sure about the approach but #500 addresses this issue.

PR #500, opened 2017-03-28
Assume USSD is a valid channel + test

smn:
This addresses #499.

### DIFF
--- a/temba/channels/views.py
+++ b/temba/channels/views.py
@@ -527,6 +527,7 @@ def channel_status_processor(request):
         cutoff = timezone.now() - timedelta(hours=1)
         send_channel = org.get_send_channel(scheme=TEL_SCHEME)
         call_channel = org.get_call_channel()
+        ussd_channel = org.get_ussd_channel(scheme=TEL_SCHEME)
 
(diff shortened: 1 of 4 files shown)


smn:
@nicpottier & @rowanseymour ready for review. Not sure about the approach.

temba/channels/views.py
```diff
@@ -527,6 +527,7 @@ def channel_status_processor(request):
         cutoff = timezone.now() - timedelta(hours=1)
         send_channel = org.get_send_channel(scheme=TEL_SCHEME)
         call_channel = org.get_call_channel()
+        ussd_channel = org.get_ussd_channel(scheme=TEL_SCHEME)
```
nicpottier: Why do we have scheme for `get_ussd_channel` anyways? Is there any scenario where that can't be `TEL`?

smn: right, not that I can think of. I'll refactor.

nicpottier: Awesome, thanks!

smn:
@nicpottier ready for review again

PR #501, opened 2017-03-29
Create a first message for when starting a USSD Flow

smn:
When starting a USSD session, the first message isn't saved and used in the Flow as the `start_msg`. This means that the first reply the Flow sends back has no `response_to` set. When USSD channels (Like Vumi and Junebug) try to send a reply they often need to refer back to the `external_id` of the `response_to` object to allow the integrations to link back to an actively running USSD session.

This PR should address that.

### DIFF
--- a/temba/flows/models.py
+++ b/temba/flows/models.py
@@ -1545,7 +1545,6 @@ class Flow(TembaModel):
                     for msg in step_msgs:
                         msgs.append(msg)
 
-            # no start msgs in ussd flows but we want the variable there
             run.start_msgs = [start_msg]
 
             runs.append(run)
--- a/temba/flows/models.py
+++ b/temba/flows/models.py
@@ -1539,16 +1539,14 @@ class Flow(TembaModel):
 
                 step = self.add_step(run, entry_rule, is_start=True, arrived_on=timezone.now())
                 if entry_rule.is_ussd():
-                    # create an empty placeholder message
-                    msg = Msg(org=self.org, contact_id=contact_id, text='', id=0)
-                    handled, step_msgs = Flow.handle_destination(entry_rule, step, run, msg, trigger_send=False)
+                    handled, step_msgs = Flow.handle_destination(entry_rule, step, run, start_msg, trigger_send=False)
 
                     # add these messages as ones that are ready to send
                     for msg in step_msgs:
                         msgs.append(msg)
 
             # no start msgs in ussd flows but we want the variable there
-            run.start_msgs = []
+            run.start_msgs = [start_msg]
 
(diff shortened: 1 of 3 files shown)


smn:
@nullcode could you have a look at this one?

smn:
@nullcode @nicpottier ready for review, this is required for all but 1 of the USSD transports we work with. The one that works is the only protocol implementation that doesn't require a `reply_to` value set.

temba/ussd/models.py
```diff
@@ -31,17 +31,18 @@ class USSDSession(ChannelSession):
     class Meta:
         proxy = True
 
-    def start_session_async(self, flow):
-        flow.start([], [self.contact], start_msg=None, restart_participants=True, session=self)
+    def start_session_async(self, flow, urn, content, date, message_id):
+        from temba.msgs.models import Msg, USSD
+        message = Msg.create_incoming(
```
nullcode: this will call msg.handle() that we don't need, because we want flow.start() to handle it

temba/flows/models.py
```diff
@@ -1539,16 +1539,13 @@ def start_ussd_flow(self, all_contact_ids, start_msg=None, extra=None, flow_star
 
                 step = self.add_step(run, entry_rule, is_start=True, arrived_on=timezone.now())
                 if entry_rule.is_ussd():
-                    # create an empty placeholder message
-                    msg = Msg(org=self.org, contact_id=contact_id, text='', id=0)
```
nullcode: I believe we still need this for push, I don't get how the tests are not failing, so I maybe wrong

temba/ussd/models.py
smn: Thanks, seeing what you meant now, switched to using `Msg.objects.create()` directly.

temba/flows/models.py
smn: Not sure what to do about this, we don't have any USSD Push channels so unable to test on a phone either. I'm relying on the tests here 😕

smn:
We've released this as `v3.0.90-praekelt` on docker hub to test it with some live USSD services and can confirm it's working for USSD pull now.

temba/flows/models.py
nicpottier: Honestly I don't think we really want to support push anyways, so I'm fine removing support for push.

nullcode:
well, push is still on my table, but don't worry about it here guys, I will make some changes on it later anyway, this should go in, so you can use pull in prod

smn:
Travis build failed due to some infrastructure issues they're having. The tests pass for me locally.

PR #508, opened 2017-03-30
Sometimes the `id` is not available from Junebug, log when that happens

smn:
Currently it just crashes and logs an `id` in the channel log which is not all that useful.
channel_event


smn:
Turns out this was a way more obscure bug and surfaced when we switch USSD backends in Junebug.

PR #518, opened 2017-04-04
Management command to download OSM extracts for relation ids

smn:
This accompanies the `import_geojson` management command. It takes OSM relation ids as positional arguments and then downloads the relevant files from a GitHub repository. It defaults to `nyaruka/posm-extracts`.

### DIFF
--- a/temba/locations/management/commands/download_geojson.py
+++ b/temba/locations/management/commands/download_geojson.py
@@ -1,0 +1,62 @@
+import os
+import requests
+import re
+
+from django.core.management.base import BaseCommand
+
+
+class Command(BaseCommand):
+    help = 'Download geojson files for OSM relation ids.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('relation_ids', nargs='+')
+        parser.add_argument('--oauth-token',
+                            dest='oauth_token',
+                            default=None,
+                            help='The OAuth token to use when authenticating '
+                                 'to GitHub. Unauthenticated requests are '
+                                 'rate limited to 60 per hour. Defaults to '
+                                 'None')
+        parser.add_argument('--repo',
+                            dest='repo',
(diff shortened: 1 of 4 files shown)


temba/locations/management/commands/download_geojson.py
```diff
@@ -0,0 +1,70 @@
...
+
+class Command(BaseCommand):
+    help = 'Download geojson files for OSM relation ids.'
+
+    def add_arguments(self, parser):
+        parser.add_argument('relation_ids', nargs='+')
+        parser.add_argument('--oauth-token',
+                            dest='oauth_token',
+                            default=None,
+                            help='The OAuth token to use when authenticating '
```
rowanseymour: max-line-length=120 ...

temba/locations/management/commands/download_geojson.py
```diff
@@ -0,0 +1,70 @@
+import os
```
rowanseymour: `from __future__ import print_function, unicode_literals`

temba/locations/management/commands/download_geojson.py
```diff
@@ -0,0 +1,70 @@
...
+import requests
+import re
```
rowanseymour: We've been switching to `regex` instead of `re`

smn:
@rowanseymour thanks, updated.

The reason I added this was so that we could automate the geojson stuff using environment variables in a docker deploy. If not of interest here I can ship it as part of the docker builds.

smn:
@nicpottier 👍  👎  ?

nicpottier:
Sorry, tried to comment on this over email but apparently didn't work.

This is definitely super cool, note that the Nyaruka extracts are definitely stale but this is a nice bootstrap. Someday when we have lots of spare time I really want to build these automatically and publish them on a fancy site, but ya, not a lot of that.

PR #520, opened 2017-04-05
history_class template tag crashes when given a Msg object.

smn:
It looks for an attribute called `recipient_count` which is on the `broadcast`, not the `Msg` itself.

### DIFF
--- a/temba/contacts/templatetags/contacts.py
+++ b/temba/contacts/templatetags/contacts.py
@@ -199,7 +199,7 @@ def history_class(item):
     if isinstance(item, Msg):
         if item.media and item.media[:6] == 'video:':
             css = '%s %s' % (css, 'video')
-        if item.direction or item.recipient_count:
+        if item.direction or (item.broadcast and item.broadcast.recipient_count > 1):
             css = '%s %s' % (css, 'msg')
         if item.status in (ERRORED, FAILED):
             css = '%s %s' % (css, 'warning')
--- a/temba/contacts/models.py
+++ b/temba/contacts/models.py
@@ -519,6 +519,7 @@ class Contact(TembaModel):
 
         msgs = Msg.objects.filter(contact=self, created_on__gte=after, created_on__lt=before)
         msgs = msgs.exclude(visibility=Msg.VISIBILITY_DELETED).select_related('channel').prefetch_related('channel_logs')
+        msgs = msgs.prefetch_related('broadcast')
 
         # we also include in the timeline purged broadcasts with a best guess at the translation used
         recipients = BroadcastRecipient.objects.filter(contact=self)
--- a/temba/contacts/tests.py
+++ b/temba/contacts/tests.py
@@ -1578,6 +1578,7 @@ class ContactTest(TembaTest):
         from temba.contacts.templatetags.contacts import activity_icon, history_class
 
         self.assertEquals('non-msg', history_class(call))
+        self.assertEquals('msg', history_class(msg))
 
         call.status = IVRCall.FAILED
         self.assertEquals('non-msg warning', history_class(call))


temba/contacts/templatetags/contacts.py
```diff
@@ -199,7 +199,7 @@ def history_class(item):
     if isinstance(item, Msg):
         if item.media and item.media[:6] == 'video:':
             css = '%s %s' % (css, 'video')
-        if item.direction or item.recipient_count:
+        if item.direction or (item.broadcast and item.broadcast.recipient_count > 1):
```
rowanseymour: Prefetch for broadcast in `Contact.get_activity(..)` ?

smn: @rowanseymour thanks for the review but I'm not sure what you're asking me here. Could you elaborate a bit? I've copied the approach taken in the [`activity_icon()` template tag](https://github.com/praekeltfoundation/rapidpro/blob/406dc784bb40b799021bcd41386be37629a46ad3/temba/contacts/templatetags/contacts.py#L163).

rowanseymour: @smn I mean make sure that messages being displayed in the contact history have broadcast prefetched so this line doesn't turn into a db hit for each message. But am still confused about how item.direction can be non-truthy....

smn: Ah, then I suspect this [recent change](https://github.com/rapidpro/rapidpro/commit/e6fc0ade5f442ccd88752b3d3cf5f9510666cf2c) is the reason. That creates an empty inbound message when starting a USSD session, the direction isn't being set there.

That would also explain why you're not seeing this error on your installs.

smn: wrt the prefetching, this stuff is called by the Contact's History SmartReadView which does `contact.get_activity()` which fetches the [`broadcast` using `select_related`](https://github.com/rapidpro/rapidpro/blob/406dc784bb40b799021bcd41386be37629a46ad3/temba/contacts/models.py#L526).

rowanseymour: Ok well go to know I'm not losing my mind! 

I think `get_activity()` there is using `select_related` for fetching purged broadcasts, but not for messages.

smn: cool, prefetching it for messages now.

smn:
@rowanseymour ready for review again, fixing the `msg.direction` issue in #521

rowanseymour:
With #521 merged you don't need this? The tag in question needs cleaned up more given that item.direction is always set - so I can close this if you want or you can make more changes.

smn:
I don't need this merged anymore but happy to leave this here if you (or someone else) wants to take up that cleanup task. Otherwise 👍  on closing it.

rowanseymour:
I'll close this for now and add it to my todo list

smn:
Just a heads up that I'm merging this into our fork as we have a few (28) records with `direction` set to an empty string. The database functions & triggers installed that deal counts prevent us from manually correcting this direction via SQL.

rowanseymour:
Ok I'm actually tidying up that whole tags file at the minute. I really think you need to fix those records because we can't guarantee that other parts of RapidPro aren't going to assume that direction=I|O and blow up if not.

smn:
yeah, just noting it here because if anyone else ran this code (as unlikely as it is) is likely going to be having the same problem in their databases.

Does removing a contact result in their messages getting removed? If not, are messages safe to delete individually?

rowanseymour:
You can delete a message without deleting the contact (we do this when we purge broadcasts), but you can't delete a contact without cascade deleting the messages.

smn:
ok that's simpler, I've reverted the merge and will remove the individual messages rather.

PR #521, opened 2017-04-05
The first Msg created for USSD Pull sessions has no direction set.

smn:
This causes template rendering problems because the `msg.direction` should always be truth-y.

### DIFF
--- a/temba/ussd/models.py
+++ b/temba/ussd/models.py
@@ -38,7 +38,7 @@ class USSDSession(ChannelSession):
             channel=self.channel, contact=self.contact, contact_urn=self.contact_urn,
             sent_on=date, session=self, msg_type=USSD, external_id=message_id,
             created_on=timezone.now(), modified_on=timezone.now(), org=self.channel.org,
-            direction=self.USSD_PULL)
+            direction=self.INCOMING)
         flow.start([], [self.contact], start_msg=message, restart_participants=True, session=self)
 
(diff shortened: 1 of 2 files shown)


temba/ussd/models.py
```diff
@@ -37,7 +37,8 @@ def start_session_async(self, flow, urn, content, date, message_id):
         message = Msg.objects.create(
             channel=self.channel, contact=self.contact, contact_urn=self.contact_urn,
             sent_on=date, session=self, msg_type=USSD, external_id=message_id,
-            created_on=timezone.now(), modified_on=timezone.now(), org=self.channel.org)
+            created_on=timezone.now(), modified_on=timezone.now(), org=self.channel.org,
+            direction=self.USSD_PULL)
```
nicpottier: Using `USSD_PULL` here is slightly weird to me, only because it REALLY matters that we only use `I` and `O` for directions because of how indexes are build and queries are optimized and my concern is that we could change the constants above to be something else and not notice that we've screwed ourselves in that respect.

How do you feel about just using `Msg.INCOMING`?

smn: I'm fine with `Msg.INCOMING`, just defaulted to `USSD_PULL` because it's the attribute that's set and used in the [`USSDSession` model](https://github.com/praekeltfoundation/rapidpro/blob/3e7abec4386518a28d683307f1f1c739e08e7616/temba/ussd/models.py#L27)

PR #527, opened 2017-04-11
Fix GeoJSON loading from directories.

smn:
This PR addresses two things:

1. Really skip when not matching a file instead of printing that we're skipping.
2. Allow loading geojson files from directories instead of assuming it is in the current directory.


smn:
Ready for review.

PR #543, opened 2017-04-27
Change contacts API v1

teehamaral:
This PR includes a change on contacts API v1 to add contact fields when they don't exist. 

We know that the API v1 is legacy now, but Surveyor app still uses it, so can happen of some field be removed and cause some error when send data to server. This PR can prevent this issue.


rowanseymour:
@teehamaral can you explain a bit more about the need here? In your previous PR you had spotted that the steps endpoint doesn't create new fields - that does seem a bug that should be fixed. The fields list passed to the steps endpoint includes the all the details required to create a new field.

This isn't the case with the contacts endpoint which only has a field key. Is there a reason you can't use the API v2 fields endpoint to ensure that fields exist? 

Is this is problem you've seen in surveyor or are you building something else on the v1 steps endpoint?

johncordeiro:
Hey @rowanseymour, this problem only occurs when users run flows on Surveyor app which contains fields that don't exist in RapidPRO, in this case, when they try to upload the data collected from a particular flow, Surveyor app responds with an unknown error and doesn't send the data, the users go crazy and make the log out, losing the data 😢

In my opinion is a necessary change even if we update Surveyor to API v2, because we don't control previous versions.

By the way, are you open to receive an update on Surveyor app to API v2?

rowanseymour:
@johncordeiro I think we can merge this because *only* old surveyor clients should be using API v1 at this point, but we won't change the API v2 contacts endpoint to behave similarly. It's preferable to create new fields in a way that can provide label and value type.

Regarding Surveyor using API v2, it's intentional that v2 doesn't have a steps endpoint yet. That's because we're planning some changes to how flows are run and how results are saved. So at this point I'd advise holding off on moving Surveyor to v2.

PR #546, opened 2017-04-28
Preparing to accept new FCM channel API key

teehamaral:
Firebase upgraded the server keys to a new version. It's possible continue to use your inherited key from the server, but they are recommending upgrading to the latest version. So this PR prepares the Channel model to accept a new key with more than 150 characters.


teehamaral:
👍

PR #573, opened 2017-05-25
Chatbase integration

teehamaral:
This PR includes an integration with Chatbase (https://chatbase.com).

Chatbase analytics helps developers monitor and fix bots. Some benefits: monitor the health of your chatbot, fix broken experiences and understand user journeys.

In the RapidPro, the Chatbase configuration is at organization settings, being able the organization to connect or not to Chatbase. If the organization is connected, when it have incoming or outgoing messages the Chatbase API is requested. So, the organization can analyze in the Chatbase panel the behavior of the bot with the users.


nicpottier:
Neat! Though really wish you had talked to us about this before starting implementing as it may have saved you some time.

The biggest issue I see here is that you are creating new database objects for every single outgoing message? First that doesn't seem necessary, seems we could be sticking the content of what needs to be sent to chatbase in some JSON that is in redis, second, you are going to destroy the main celery task queue with these events at scale which is going to cause real problems. This should probably be using the celery msg queue instead and at the lowest priority so that chatbase events only get sent when all msgs have been handled.

It looks like the chatbase API support bulk updates, so that seems like maybe the best approach to log these events. Perhaps something similar to how we log to librato where a background thread collects events and then fires them off as a batch every now and then. (it's a bit trickier as these need to be org specific but probably still doable)

Shouldn't incoming messages be logged through chatbase as well? Doesn't seem like this is a full enough record to really provide the best analytics. There seem to be a lot of other fields that would be interesting to log, such as whether the message was handled by a flow or not, which flow etc..

teehamaral:
Hey @nicpottier, thanks for your quick feedback. So, could you say, please, if I understood about your thinking, checking the below items?
1 - We don't need to save the requests to Chatbase on database to display on log page;
2 - The Chatbase request must be at msgs queue on Celery with low priority;
3 - Try using the batch process to send requests, but I didn't can find the API on Chatbase to do that until now :(
4 - Didn't we add to send to Chatbase the incoming messages?
We also think that would be necessary to send more data to Chatbase like you said but on Chatbase documentation (https://chatbase.com/documentation/generic) there just some parameters about message data.

nicpottier:
Saw the bulk / new stuff by glancing at the docs for this node package:  
   https://www.npmjs.com/package/@google/chatbase

Yes, no db objects please!

But yes, sounds like you understood the approach there.

teehamaral:
Could you check this solution, please? I'll type the tests after the code is reviewed.

temba/orgs/tasks.py
```diff
@@ -45,3 +48,18 @@ def calculate_credit_caches():  # pragma: needs cover
...
+
+@task(track_started=True, name='send_chatbase_logs')
+def send_chatbase_logs():  # pragma: needs cover
+    """
+    Send messages logs in batch to Chatbase
+    """
+    from temba.orgs.models import CHATBASE_API_KEY
+
+    for org in Org.objects.filter(config__icontains=CHATBASE_API_KEY):
+        cache_keys = cache.keys('org:%d:cache:chatbase_log:*' % org.id)
```
rowanseymour: Redis KEYS command should generally not be used in production because of performance issues (https://redis.io/commands/keys)

Better that chatbase logs are stored in a single Redis list which acts as a queue.

temba/channels/models.py
```diff
@@ -1329,6 +1330,12 @@ def success(cls, channel, msg, msg_status, start, external_id=None, event=None,
                                       response_status=event.status_code,
                                       request_time=request_time_ms)
 
+            # Send data to Chatbase API
+            if hasattr(msg, 'is_org_connected_to_chatbase'):
+                Org.register_chatbase_log(api_key=msg.chatbase_api_key, version=msg.chatbase_version, org_id=msg.org,
```
rowanseymour: We don't need to store the chatbase credentials on the message or the log - we can get these from the org when sending the logs

rowanseymour: Also let's call this something like `queue_chatbase_log` rather than `register_chatbase_log`

templates/chatbase/chatbase_read.haml
```diff
@@ -0,0 +1,36 @@
+{% extends "smartmin/read.html" %}
```
rowanseymour: Is this file being used any more?

temba/orgs/tasks.py
teehamaral: So, just separated by org? Example: `org:%d:cache:chatbase_log`

rowanseymour: I'm thinking it would be ok to have single queue for everyone, at least for now. Makes it easier to make sure we're not leaving stuff in Redis if an org disables ChatBase support.

@nicpottier thoughts?

nicpottier: Ya, agreed that it seems having a single queue should be ok for this.

teehamaral: Currently we are changing the value to `not_handled` parameter after the messages are handled, it's happening on Redis key before sending to Chatbase. So, working this way (just one Redis key to each organization), we have a problem, iterate in messages array to change this parameter, and it's very important to Chatbase like the documentation says. Could you give an idea about resolving that? Thanks.

rowanseymour: I'm not quite following. I think we should have a single queue (across all orgs) for chatbase events - and as these are sent to chatbase, they are thrown away.

teehamaral:
Could you review? We include flow name as `intent` in Chatbase payload.

rowanseymour:
Hi @teehamaral we haven't forgotten about this - we've just started discussing about how this fits with other analytics services that users might want to use, and how we can make this generic so it's easier to support different services.

teehamaral:
Hi @rowanseymour, would you have any position about?

rowanseymour:
Hi @teehamaral. Apologies for the delay getting back to but we have now finally figured out what we think is the best way forward for this stuff.

1. It's ok for this not to be a generic solution just yet. If someone requests support for another analytics service than we can revisit this. So no changes to this PR required regarding that.
2. Let's go for simplicity. We can send requests to the Chatbase API inline with receiving and sending messages in RapidPro. 
    - This means we don't need additional redis queuing for chatbase logs
    - The current queueing for sending messages is fairly balanced between orgs so we don't have to replicate that for sending chatbase logs

That make sense?

teehamaral:
No problem about the delay @rowanseymour.

Yeah, I did it in the first place but @nicpottier said that would be a better solution sending a JSON to Redis (https://github.com/rapidpro/rapidpro/pull/573#issuecomment-304037585) instead. I think that using Redis is more reliable than single requests, using Redis we also can send messages in batches.

rowanseymour:
@teehamaral I thought the first iteration of this was creating db objects? I know it's a pain to keep changing this but we'd really like to keep this as simple as possible for now and not introduce additional queues and tasks. Sending the chatbase logs inline with sending messages means we'll get fair balancing between orgs for free.

teehamaral:
Hi @rowanseymour, could you review please?

temba/channels/models.py
```diff
@@ -1319,6 +1321,13 @@ def success(cls, channel, msg, msg_status, start, external_id=None, event=None,
                                       response_status=event.status_code,
                                       request_time=request_time_ms)
 
+            # Send data to Chatbase API
+            if hasattr(msg, 'is_org_connected_to_chatbase'):
+                on_transaction_commit(lambda: send_chatbase_logs.apply_async(args=(msg.chatbase_api_key,
```
rowanseymour: Is there a reason for doing this async? `ChannelLog.success` is going to be called from a send message task which is fairly balanced across orgs.

rowanseymour:
@teehamaral can we get some screenshots of the new UI stuff

temba/msgs/models.py
```diff
@@ -831,9 +834,22 @@ def process_message(cls, msg):
...
         # if this is an inbox message, increment our unread inbox count
         if msg.msg_type == INBOX:
             msg.org.increment_unread_msg_count(UNREAD_INBOX_MSGS)
+        elif msg.msg_type == FLOW:
+            chatbase_not_handled = False
+
+        # Registering data to send to Chatbase API later
+        org = msg.org
+        if org.is_connected_to_chatbase():
+            on_transaction_commit(lambda: send_chatbase_logs.apply_async(
```
teehamaral: @rowanseymour change to not call async here too?

teehamaral:
Image 1:
screen shot 2017-08-15 at 11 37 43

Image 2:
screen shot 2017-08-15 at 11 35 20

Image 3:
screen shot 2017-08-15 at 11 37 15

Image 4:
screen shot 2017-08-15 at 11 37 25

temba/msgs/models.py
rowanseymour: yeah

teehamaral:
@rowanseymour I just changed the call to chatbase to be not async, could you review now please?

temba/msgs/models.py
```diff
@@ -844,6 +859,36 @@ def process_message(cls, msg):
         analytics.gauge('temba.channel_handling_latency', (msg.modified_on - msg.created_on).total_seconds())
 
     @classmethod
+    def send_chatbase_logs(cls, chatbase_api_key, chatbase_api_version, channel_name, text, contact_id, log_type,
```
rowanseymour: `send_chatbase_log` since this only ever sends a single log now

temba/msgs/models.py
```diff
@@ -831,9 +835,20 @@ def process_message(cls, msg):
...
+
         # if this is an inbox message, increment our unread inbox count
         if msg.msg_type == INBOX:
             msg.org.increment_unread_msg_count(UNREAD_INBOX_MSGS)
+        elif msg.msg_type == FLOW:
+            chatbase_not_handled = False
+
+        # Registering data to send to Chatbase API later
+        org = msg.org
+        if org.is_connected_to_chatbase():
```
rowanseymour: Lets re-organize this to be more efficient as it's currently calling msg.as_task_json() twice and org.config_json() several times

teehamaral: I just updated it.

rowanseymour: `org.is_connected_to_chatbase` parses the org config so if we're doing that, we might as well take the chatbase credentials from the org config, and skip calling `msg.as_task_json()` altogether, no?

templates/orgs/org_chatbase.haml
```diff
@@ -0,0 +1,64 @@
...
+  -if object.is_connected_to_chatbase
+    -trans "Connected to Chatbase Account"
+    {{chatbase_agent_name}}.
+  -else
+    -trans "No Chatbase account connected."
+
+-block pre-form
+  -if not object.is_connected_to_chatbase
+    -blocktrans
+      Adding a Chatbase account will allow you to analyse and receive suggestions for how to “fix” bot experiences
```
rowanseymour: I think replace `"fix"` with `improve`

templates/orgs/org_chatbase.haml
```diff
@@ -0,0 +1,64 @@
...
+  -if object.is_connected_to_chatbase and request.META.HTTP_X_FORMAX
+    %p
+      -trans "Your account is connected to the Chatbase account"
+      {{chatbase_agent_name}}.
+
+    %p
+      -trans "If you no longer want it connected, you can"
+      %a{href:'javascript:confirmChatbaseDisconnect();'}
+        -trans "disconnect"
+      -trans "your Chatbase account. Doing so will interrupt the data send to Chatbase service."
```
nicpottier: `data sent to Chatbase.`

templates/orgs/org_chatbase.haml
```diff
@@ -0,0 +1,64 @@
...
+    .disconnect-chatbase.hide
+      .title
+        -trans "Disconnect Chatbase"
+      .body
+        -blocktrans
+          This will disconnect your Chatbase account. Are you sure you want to proceed?
+    %a#disconnect-chatbase-form.posterize{href:'{% url "orgs.org_chatbase" %}?disconnect=true'}
+  -else
+    -trans "You can find your API Key by clicking "
+    %a{href:'https://chatbase-area120.appspot.com/agents', target:'_new'}><
```
nicpottier: Maybe put this in the help for the API key?

teehamaral:
@nicpottier @rowanseymour could you review now, please?

temba/msgs/models.py
```diff
@@ -831,9 +835,21 @@ def process_message(cls, msg):
...
+
         # if this is an inbox message, increment our unread inbox count
         if msg.msg_type == INBOX:
             msg.org.increment_unread_msg_count(UNREAD_INBOX_MSGS)
+        elif msg.msg_type == FLOW:
+            chatbase_not_handled = False
+
+        # Sending data to Chatbase API
+        org = msg.org
+        if org.is_connected_to_chatbase():
```
rowanseymour: `is_connected_to_chatbase` calls `config_json()` and then it's immediately called again. Maybe instead of having a `is_connected_to_chatbase` method, you could have a `get_chatbase_credentials` which returns the key/version or None - that way you can use it to both 1) determine if org is configured for chatbase, and 2) get the credentials

teehamaral:
I have updated @rowanseymour

temba/msgs/models.py
```diff
@@ -831,9 +835,21 @@ def process_message(cls, msg):
...
         # if this is an inbox message, increment our unread inbox count
         if msg.msg_type == INBOX:
             msg.org.increment_unread_msg_count(UNREAD_INBOX_MSGS)
+        elif msg.msg_type == FLOW:
+            chatbase_not_handled = False
+
+        # Sending data to Chatbase API
+        chatbase_credentials = msg.org.get_chatbase_credentials()
+        if chatbase_credentials:
+            cls.send_chatbase_log(chatbase_credentials.get(CHATBASE_API_KEY.lower()),
```
rowanseymour: This is still a bit more complex than it needs to be. How about instead of `get_chatbase_credentials` packing things into a dict, and then having to extract them here... just return a tuple of (key, version)

teehamaral:
@rowanseymour I think that's ready to merge 😀

rowanseymour:
Yup thanks @teehamaral!

teehamaral:
Thanks for your help with that @rowanseymour

PR #576, opened 2017-05-31
Add header to webhook on flow step

teehamaral:
This PR includes header to webhook on flow steps (actionsets and rulesets) #146.


templates/partials/split_editor.haml
```diff
@@ -113,6 +113,18 @@
 
         %select{ng-model:"formData.webhook_action",ng-options:"method for method in methods",style:"height:34px;width:95px",class:"method",select2:""}
         %input{ng-model:"formData.webhook",placeholder:"http://yourwebsite.com/endpoint",auto-complete:"",ng-required:"true",type:"url",class:"url"}
+
+        .webhook-headers
+          .modal-webhook-header-pair
+            .modal-webhook-header.modal-webhook-header-key
+              .message
+                -trans "Header key"
```
nicpottier: "Header Name"

templates/partials/split_editor.haml
```diff
@@ -113,6 +113,18 @@
...
+              %input{type:"text",name:"webhook-key",ng-model:"formData.webhook_header_key",ng-change:"updateWebhookHeader()"}
+            .modal-webhook-header.modal-webhook-header-value
+              .message
+                -trans "Header value"
```
nicpottier: "Header Value"

static/coffee/flows/controllers.coffee
```diff
@@ -1640,6 +1650,7 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
         ruleset.config =
           webhook: formData.webhook
           webhook_action: formData.webhook_action
+          webhook_header: formData.webhook_header
```
nicpottier: This can be more than one header ya? So should rename this to `webhook_headers` everywhere.

teehamaral: In this moment just one item, but I'll try to add to support more than one.

teehamaral:
Ruleset:
screen shot 2017-06-01 at 14 12 56
Actionset:
screen shot 2017-06-01 at 14 13 11

teehamaral:
Updated to support more than one item in header

ActionSet:
screen shot 2017-06-04 at 19 04 46
RuleSet:
screen shot 2017-06-04 at 19 05 24

nicpottier:
It'd be nice to figure out a UI here that wasn't so heavy. Instead of having two lines, instead having something that expanded or hid or something.

teehamaral:
This two lines was add to exemplify, but just link "Add new header item" is displayed when don't exist items on header.

ericnewcomer:
A few comments on the UI:

* The header label should only show for the first header added, no reason to repeat that for each header
* You should be able to hover over a header and have an "x" to the right to delete it (consistent with how the rule editor works)

I'm okay with a link for adding new headers.. how about just "Add header" for the language. Agree with Nic.. headers should be hidden by default with an option to expand to see this UI.

teehamaral:
Can you review this UI, please?

screen shot 2017-06-06 at 10 04 29
screen shot 2017-06-06 at 10 04 57
screen shot 2017-06-06 at 10 05 38
screen shot 2017-06-06 at 10 16 04

nicpottier:
Sweet! Let's get rid of "Additional Options" after you click it (it's ok if they can't make it go away without reopening the dialog). But otherwise I think we are set.

@rowanseymour want to make one last pass?

teehamaral:
Review please

PR #588, opened 2017-06-26
Flow configuration improvements for USSD

nullcode:
Includes:

- Options to enable/disable USSD Push for a flow
- Add channel trigger behaviour from the start new flow modal
- USSD Interrupted state improvements to respect disabled USSD push configuration (cannot connect nodes to the interrupt rule that would send messages through the USSD channel.

Some screenshot that represent the new changes:
- Create Flow
screen shot 2017-06-26 at 17 43 24

- Update Flow
screen shot 2017-06-26 at 17 43 44

- Invalid Connection
screen shot 2017-06-26 at 17 44 03


nullcode:
ready for review @nicpottier @ericnewcomer

nicpottier:
Just out of curiosity is this being driven by customer requirements?

nullcode:
Comes from @ewheeler , so he might know more, but this definitely improves the UX a lot, Push disabled by default, ability to add the USSD code from the create page are features that we were talking about at RapidCon.

RichardHeslopZA:
Regarding the interrupt/resume functionality, this does not appear to be working in our environments. We are running v3.0.136.

We have linked the connector from the 'Interrupted' category back to the same node (see below). We have also set the 'Expire inactive contacts' value for the flow to the shortest option (5 minutes).

capture

Has the USSD interrupt/resume functionality been tested & confirmed to be functioning reliably as expected? Is this PR expected to resolve known issues with the interrupt/resume functionality?

Thanks

nullcode:
@RichardHeslopZA I believe what you are trying to do here is not working, because you are using a USSD Pull channel. If you don't have push capability you can't start (or resume) sessions from the server. This PR fixes this from a UX point of view so you can't make these kind of connections that wouldn't work in the first place. So again, for this to work you will need a USSD Push capable channel.

@nicpottier @ericnewcomer can we get a review here?

RichardHeslopZA:
Thanks @nullcode . 

I want to confirm I understand correctly: you are saying that the interrupt/resume functionality in USSD flows was not developed for use with USSD Pull channels?

nullcode:
@RichardHeslopZA I was talking about the particular connection (self-loop) that you posted on your screenshot. With Pull you can only make connections to non-messaging nodes (eg: action that will put the user to a group called "Failed"). If done otherwise the new PR will prevent you (screenshot 3) to make connections.

RichardHeslopZA:
Thanks @nullcode .

Would you be able to share a screenshot of a flow that provides an example of how we to correctly use the interrupt/resume feature on a Pull USSD session?

nullcode:
@RichardHeslopZA here is a test case with an interrupted example, you can see in the simulator how it works (you can actually test the interrupt branches by sending the special interrupt keyword to the simulator).

screen shot 2017-07-05 at 13 55 28

I want to emphasise that after this PR gets merged you don't have to worry about how to use the interrupts, because the flow editor will not let you make a broken flow, so you can only connect pieces together that will work for your channel.

temba/flows/views.py
```diff
@@ -578,6 +620,15 @@ def __init__(self, user, *args, **kwargs):
...
                     self.fields[Flow.CONTACT_CREATION] = contact_creation
+                elif self.instance.flow_type == Flow.USSD:
+                    ussd_trigger = Trigger.objects.filter(
+                        org=self.instance.org, flow=self.instance, is_archived=False, groups=None,
+                        trigger_type=Trigger.TYPE_USSD_PULL
+                    ).first()
+                    if ussd_trigger:
+                        self.fields['ussd_trigger'].initial = ussd_trigger.keyword
+                        self.fields['ussd_trigger'].widget.attrs['readonly'] = True
+                        self.fields['ussd_trigger'].help_text = 'To change the USSD trigger code, go to Triggers'
```
ericnewcomer: This smells a bit, can you explain what's going on here?

temba/flows/views.py
```diff
@@ -948,14 +1009,17 @@ def get_gear_links(self):
             links = []
             flow = self.get_object()
 
-            if flow.flow_type not in [Flow.SURVEY, Flow.USSD] \
+            if flow.flow_type != Flow.SURVEY \
```
ericnewcomer: Would much rather see a flow.allows_start() or somesuch

static/coffee/flows/services.coffee
```diff
@@ -741,17 +741,49 @@ app.factory 'Flow', ['$rootScope', '$window', '$http', '$timeout', '$interval',
...
+          @updateDestination(source, null)
+
     # check if a potential connection would result in an invalid loop
     detectLoop: (nodeId, targetId, path=[]) ->
 
       node = @getNode(targetId)
 
       # can't go back on ourselves
-      if nodeId == targetId and not @isUssdRuleset(node)
+      if nodeId == targetId and (not @isUssdRuleset(node) or @isUssdRuleset(node) and not window.ussd_push_enabled)
```
ericnewcomer: These conditions are getting a bit crazy. Can you simplify and/or make this a method?

temba/flows/views.py
nullcode: The idea here is that the ussd trigger input here is simply adding a trigger (if you haven't added one yet) or just show it. Updating, deleting would be difficult (eg: cases when you have more triggers for the same flow, etc.), so I decided to keep it simple here and if you want to do updates or deletes you should go to the Trigger section and do it there.

ericnewcomer: If it can't be edited here as well, I'd argue for omitting it completely and letting this happen on the trigger page.

nullcode: @ericnewcomer okay, should I remove it from create & update too or only from update ?

ericnewcomer: I see two options:

* Remove it from both
* Make updates work

Honestly, somewhat indifferent to which approach you take for this, just feel it should be consistent.

nullcode: okay, thanks!

nullcode:
@ericnewcomer I decided to go with implementing update and delete from the update flow modal mainly because I find it useful to create a ussd trigger from the create flow modal.
Had a PIP mismatch issue (I think a version of pyexcel-io actually got deleted from PIP :) and also on this PR Travis seems to be stuck, it ran green on my branch.
Otherwise it's ready.

RichardHeslopZA:
@nullcode 

2 questions:
1) What is the reason for limiting the interrupt functionality to USSD Push channels? (We are familiar with other USSD messaging systems that support interrupt/resume for USSD Pull channels. Praekelt's USSD based projects all make exclusive use of USSD Pull)
2) The screenshot explains how the interrupt is handled (i.e. in this case you have removed from groups and added a label). Can you provide a screenshot that shows how the resume portion would be implemented? For example, if a user's session timed-out & he/she redials, we would want to re-insert them in the flow at the point where they were interrupted.

Thanks!

nullcode:
@RichardHeslopZA 
1. It's only limited to non-messaging nodes since the session was interrupted and you can't re-initiate sessions with ussd pull. Unless I'm mistaken the resume interrupt is something that you guys implemented for your own gateways and is not a standard. @smn mentioned it in #593.
2. there's no option to resume a session when using ussd pull, the user would have to restart the flow from the beginning at this point.

nullcode:
@ericnewcomer is everything done here?

nullcode:
@nicpottier @ericnewcomer can this be merged now ?

ericnewcomer:
@nullcode makes sense, can we have updated screenshots?

nullcode:
@ericnewcomer there are no changes in the screenshots, only a minimal on the update screen (now you can update the shortcode from there)
screen shot 2017-08-28 at 11 13 03

rowanseymour:
Closing this as it's 18 months stale

PR #591, opened 2017-06-30
Use USSD session id if available

smn:
We're coming across a bug in how the end of a USSD session is determined. Sometimes we're seeing a session being closed on a step that is not the end of a flow.

The current implementation uses the ordinal date to create a session identifier, this lead to erratic behaviour while doing field testing. We're switching that out to simply use the `from` address as a GSM phone is only capable of running one USSD session at any given moment.

While diagnosing this we're trying to make it easier for RapidPro to understand when a session should start and end when using USSD. For aggregator / MNO junebug integrations that do provide some kind of session identifier we are now relaying that via the API for RapidPro to use.

Whether or not this session_id is available really depends on the integration + protocol to the networks so it's not possible for us to guarantee its existence.


erikh360:
@nicpottier can you have a look at this PR please?

nicpottier:
This seems to have a bit of overlap with: 
   https://github.com/rapidpro/rapidpro/pull/613

Do you want to combine these into one? Didn't look at the graph to see if that branch is perhaps based on this.

temba/ussd/models.py
```diff
@@ -117,8 +117,19 @@ def handle_incoming(cls, channel, urn, date, external_id, contact=None, message_
         # check if there's an initiated PUSH session
         session = cls.objects.get_initiated_push_session(contact)
 
+        created = False
         if not session:
-            session, created = cls.objects.update_or_create(external_id=external_id, defaults=defaults)
+            try:
+                session = cls.objects.select_for_update().exclude(status__in=ChannelSession.DONE)\
```
nicpottier: I'm slightly apprehensive about the `select_for_update` here. We are trying to prevent a race? Generally we like depending on redis locks over db locks just because they are a bit easier to reason about, in this case I would expect a lock on the contact being the right thing. (for example when "handling" an incoming message in a flow we lock on the contact so only one thread in the system can be moving a contact forward in a flow)

erikh360: @nicpottier I see the `update_or_create` function we replaced also does a `select_for_update` so it's not adding a new lock: https://github.com/django/django/blob/master/django/db/models/query.py#L469

erikh360:
@nicpottier I combined the other PR with this one, can you have another look please?

nicpottier:
thanks guys!

PR #600, opened 2017-07-20
Quick Replies

nicpottier:
We've talked internally and we'd like to bring this forward but we'll need some changes on this implementation. We want to launch with this implemented on Telegram, Twitter and Facebook.

Notes:
 * On Twitter these are quick replies
 * On Telegram these are inline keyboards
 * On Facebook these are quick replies
 * For now, the user should only specify the label (and that is also the payload). The implementation of the handlers should create a message with the payload however. (so that later we can add the option of the user specifying both a label and payload)
 * Labels need to be localized (translated), they can also be expressions
 * Msg should have a new field called `metadata` which is a nullable text (json) field
 * Since both Twitter and Facebook use quick reply as a way of talking about these, we should too. So the field in the `metadata` should be `quick_replies`
 * For the time being we want to just add simple quick replies, we'll expand on this concept later.

I'll add more inline code review comments as well. Thanks Ihlasoft for pushing us on this, will be good to have this in there and the impact won't be too large for us to roll this into the new engine.


nicpottier:
@teehamaral @johncordeiro

karma/flows/test_controllers.coffee
```diff
@@ -765,3 +765,51 @@ describe 'Controllers:', ->
...
+
+          
+     it 'should generate json quick replies to send', ->
+      loadFavoritesFlow()
+
+      actionset = flowService.flow.action_sets[0]
+      action = actionset.actions[0]
+    
+      json_quick_responses = [{
+                    'title':'Quick reply',
```
nicpottier: let's use `label` and payload for these. Leave `content_type` out for  the time being.

temba/contacts/models.py
```diff
@@ -1807,7 +1807,7 @@ def raw_tel(self):
         if tel:
             return tel.path
 
-    def send(self, text, user, trigger_send=True, response_to=None, message_context=None, session=None,
+    def send(self, text, user, trigger_send=True, response_to=None, message_context=None, session=None, additional_params=None,
```
nicpottier: `metadata` instead of `additional_params`

karma/flows/test_controllers.coffee
```diff
@@ -765,3 +765,51 @@ describe 'Controllers:', ->
...
+                    'payload':'Test quick reply is ok',
+                    'content_type':'text'
+                }]
+
+      editAction actionset, action, (scope) ->
+        scope.AddNewQuickResponse()
```
ericnewcomer: This shouldn't be capitalized. Why not pass the initialization to this function instead of setting them below?

karma/flows/test_controllers.coffee
```diff
@@ -765,3 +765,51 @@ describe 'Controllers:', ->
...
+      action = actionset.actions[0]
+    
+      json_buttons_reply = [{
+                    'title':'url tittle reply',
+                    'url':'url.com',
+                    'type':'web_url'
+                }]
+
+      editAction actionset, action, (scope) ->
+        scope.AddNewButtonReply()
```
ericnewcomer: Same here.

temba/flows/models.py
```diff
@@ -5024,11 +5024,15 @@ class ReplyAction(Action):
     MSG_TYPE = None
     MEDIA = 'media'
     SEND_ALL = 'send_all'
+    QUICK_RESPONSES = 'quick_responses'
```
nicpottier: `quick_reply`

temba/flows/models.py
```diff
@@ -5024,11 +5024,15 @@ class ReplyAction(Action):
...
+    BUTTONS_REPLY = 'buttons_reply'
```
nicpottier: Nix this for now.

temba/msgs/migrations/0101_msg_additionalparams.py
```diff
@@ -0,0 +1,20 @@
...
+        ('msgs', '0100_auto_20170614_0915'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='msg',
+            name='additionalParams',
+            field=models.TextField(help_text='The additional params for any type msgs', null=True),
+        ),
+    ]
```
nicpottier: Get rid of this migration

temba/msgs/migrations/0102_auto_20170717_1435.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('msgs', '0101_msg_additionalparams'),
+    ]
+
+    operations = [
+        migrations.RenameField(
+            model_name='msg',
+            old_name='additionalParams',
+            new_name='additional_params',
```
nicpottier: `metadata`

temba/msgs/models.py
```diff
@@ -724,6 +725,8 @@ class Msg(models.Model):
     attachments = ArrayField(models.URLField(max_length=255), null=True,
                              help_text=_("The media attachments on this message if any"))
 
+    additional_params = models.TextField(null=True, help_text=_("The additional params for any type msgs")) # created to use to save any additional params
```
nicpottier: `metadata`

temba/msgs/models.py
```diff
@@ -960,13 +963,18 @@ def mark_sent(cls, r, msg, status, external_id=None):
             Msg.objects.filter(id=msg.id).update(status=status, sent_on=msg.sent_on)
 
     def as_json(self):
-        return dict(direction=self.direction,
+        data = dict(direction=self.direction,
```
nicpottier: Let's move away from using `data` as a variable, it isn't descriptive at all. `msg_json` would be better.

temba/msgs/models.py
```diff
@@ -960,13 +963,18 @@ def mark_sent(cls, r, msg, status, external_id=None):
...
                     text=self.text,
                     id=self.id,
                     attachments=self.attachments,
                     created_on=self.created_on.strftime('%x %X'),
                     model="msg")
 
+        if self.additional_params: #if msg have additional params return this in json
+            data.update(dict(additional_params=self.additional_params))
```
nicpottier: seems like:

```python
msg_json['metadata'] = self.metadata
```

Would be simpler

temba/msgs/models.py
```diff
@@ -1181,6 +1189,9 @@ def as_task_json(self):
 
         if self.contact_urn.auth:
             data.update(dict(auth=self.contact_urn.auth))
+            
+        if self.additional_params:
+            data.update(dict(additional_params=self.additional_params))
```
nicpottier: Same here, simple setting of the parameter would be good. (same above with auth)

templates/partials/attachment_viewer.haml
```diff
@@ -10,3 +10,40 @@
...
+
+          .row{ng-repeat:"response in action.quick_responses"}
+
+            .show-quick-response-value  
+              -trans "[[response.title]]"
+
+            .show-quick-response-value 
+              -trans "[[response.payload]]"
+
+        .show-quick-response.attachment-media{ng-if:'action.buttons_reply.length > 0', controls:'', class:""}
```
nicpottier: No button reply in this go, just quick replies, so remove this.

static/coffee/flows/controllers.coffee
```diff
@@ -1805,6 +1825,31 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
+        url:"",
+        type:"web_url"
+      })
+      return
+
+  $scope.RemoveElementArray = (a, index) ->
+    a.splice(index,1)
+    if a.length == 0
+      $scope.container_operation_visible = true
+
```
ericnewcomer: These methods should all be lower case

static/coffee/simulator.coffee
```diff
@@ -42,6 +41,20 @@ window.updateSimulator = (data) ->
...
 
+    quick_responses = null
+    buttons_reply = null
+
+    if msg.additional_params
+      params = JSON.parse(msg.additional_params)
+      if params.quick_responses[0]
+        model = 'ilog'
+        quick_responses = "<div id='quick-response-content'>"
+        for QR in params.quick_responses
```
ericnewcomer: `for reply in params.quick_replies..`

static/coffee/flows/controllers.coffee
```diff
@@ -1859,6 +1905,9 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
       $scope.action.msg = {}
     $scope.action.msg[$scope.base_language] = message
 
+    $scope.action.quick_responses = $scope.actions_quick_responses
+    $scope.action.buttons_reply = $scope.actions_buttons_reply
+
```
ericnewcomer: Only quick replies, and lets use quick_replies here instead of responses for consistency.

static/coffee/simulator.coffee
```diff
@@ -62,24 +75,33 @@ window.updateSimulator = (data) ->
...
-
-
     ele = "<div class=\"" + model + " " + level + " " + direction + " " + ussd
     if media_type
       ele += " media-msg"
+
     ele += "\">"
-    ele += msg.text
+    if quick_responses
+      console.log(quick_responses)
```
ericnewcomer: debug

ericnewcomer:
Would be nice to see some screenshots showing how the UI currently looks as well.

teehamaral:
@ericnewcomer it's like this now, but we are doing the improvements.

screen shot 2017-07-25 at 11 13 02
screen shot 2017-07-25 at 11 14 01
screen shot 2017-07-25 at 11 15 34

vctrferreira:
So, we have a problem, when use twitter API to send quick replies they need a recipient id to identify user and this parameter is not  saved, I think that save this params but needs mage server features.
I don't know how much this is viable because mage will go discontinued, what is the best way to get this id? The main problem is old users, how to get this id for olds users?

My solution is send request to twitter lookup user (https://dev.twitter.com/rest/reference/get/users/lookup) when send first msgs and save return recipient id in contact, but I don't know if this is the best solutions, wait will for you responses

nicpottier:
@rowan I think you were thinking about this as well no?

rowanseymour:
@nicpottier yeah we talked about this previously because it seemed for a time that the activity API was only going to send us Twitter user ids instead of screennames. So from what I can tell from the Twitter API docs, they're replacing their old DM REST endpoints with new ones which add the quick reply features... but require using user ids.

We need to start storing these when we receive DMs, but what are they? A field on URN? A different URN. The actual URN, and then screenname is only a display part of the URN? Need to discuss.

nicpottier:
We chatted about this internally and this is really a common problem across the three URN types we are talking about here (Telegram, Twitter and Facebook). All are now using numerical user-ids but also have the concept of a screen name, handle or username which can change but is the more natural way of referring to someone.

As this is a property on a URN and not on a contact (as a contact could have more than one Facebook URN associated with it) this belongs on URN.

Ideally we'd like to keep URNs pure in that they still fit within the URN spec and can be represented as strings. As such we propose using fragments to represent the display name for URNs which support them.

So while a current Twitter URN might be:
  `twitter:nicpottier`

We would migrate these to using ids instead and have the handle be a fragment, so they would look like this instead:
  `twitter:5818584124344#nicpottier`

The fragment is optional but can be used to display something nicer to end users in the UI. As fragments as defined in the URN spec are not part of the URN identity, this also maintains the contract that someone can change their screenname without affecting their actual URN.

So we'll make the following changes:
 * Add support for fragments / display on URN and ContactURN
 * Create a migration for current Twitter URNs to look up the Twitter ID and set it along with the appropriate display.
 * Update the sending of Twitter messages to use Twitter IDs instead of Twitter handles.

I'll open a PR for this today and try to get it ready for review by EOD.

nicpottier:
Tracking URN display attributes here:
  https://github.com/rapidpro/rapidpro/pull/608

vctrferreira:
@nicpottier what will happen with contacts inserteds by mage server this will need a feature in mage?

teehamaral:
Hello guys, can you review? I'll complete the tests and coverage after all reviews

temba/api/v2/views.py
```diff
@@ -2208,7 +2209,8 @@ class MessagesEndpoint(ListAPIMixin, BaseAPIView):
                 "status": "wired",
                 "visibility": "visible",
                 "text": "How are you?",
-                "media": "wav:http://domain.com/recording.wav"
+                "media": "wav:http://domain.com/recording.wav",
+                "metadta": {}
```
rowanseymour: typo

temba/channels/models.py
```diff
@@ -146,6 +146,11 @@ def has_attachment_support(self, channel):
         """
         return self.attachment_support
 
+    def get_quick_replies(self, **kwargs):
```
rowanseymour: Does this need to be on channel type base? Isn't it only ever called from each supporting channel's `send` method?

media/test_flows/flow_quick_replies.json
```diff
@@ -0,0 +1,105 @@
...
+                "title": "Op\u00e7\u00e3o 1"
+              },
+              {
+                "payload": "2"
+              }
+            ],
+            "eng": [
+              {
+                "payload": "1",
+                "title": "Option 1"
```
rowanseymour: I thought the plan was for implementing replies as label==payload for now? I.e. the text which the user selects is the thing we send back to rapidpro

temba/channels/types/telegram/type.py
```diff
@@ -46,11 +47,37 @@ def deactivate(self, channel):
...
+        quick_replies = metadata.get('quick_replies', None)
+        url_buttons = metadata.get('url_buttons', None)
+        replies = []
+        keyboard_type = None
+
+        if quick_replies:
+            for reply in quick_replies:
+                replies.append([dict(text=reply.get('title'), callback_data=reply.get('payload'))])
+            keyboard_type = 'keyboard'
+        elif url_buttons:
```
nicpottier: What are these url buttons?

media/test_flows/flow_quick_replies.json
nicpottier: Yes, that's right.

teehamaral: Currently, we prepared the code to receive payload as response back to RapidPro. So why wouldn't be it a good practice to do that? We think that has utility when, for example, you want to add emojis to display to the user and to send to RapidPro something different.

nicpottier: For this first release we want them to be the same, can always expand upon it later.

teehamaral: Ok, but is there any concern about it in architecture terms?

nicpottier: I don't think so, but for now we want them the same.

temba/channels/types/telegram/type.py
teehamaral: URL buttons are accepted by all channels that support quick replies, so we think that could implement them as well, and a good example for that, would be when the flow of the U-Report shares a link to poll results with his users, in this case, could use an URL button.

21015445_10211976244660395_2029988703_n

nicpottier: Guys, please, we made it clear what we were willing to accept in this PR and we explicitly said  quick replies only and with only the payload. Don't act as if this is new. Let's do this in steps as we first asked.

temba/channels/models.py
teehamaral: In the future, other channels may support quick replies, so I thought that we already could implement this way, furthermore the methods to get quick replies would need to be static.

teehamaral:
I have removed the `payload` key and URL buttons, could you review please?

temba/channels/models.py
rowanseymour: Yeah but isn't it only ever called from the channel type implementation's `send` method? I'm not getting why it needs to be defined on the `ChannelType` base

temba/channels/tests.py
```diff
@@ -7608,6 +7608,43 @@ def test_send_media(self):
             self.assertEqual(mock.call_args[0][1]['caption'], "Test message")
             self.assertEqual(mock.call_args[0][1]['chat_id'], "1234")
 
+    def test_send_quick_replies(self):
+        metadata = """
+        {
+            "quick_replies": [
+                {
+                    "payload": "yes",
+                    "title": "Yes"
```
rowanseymour: Aren't these supposed to be a single value now?

static/less/simulator.less
```diff
@@ -272,3 +272,46 @@
     display: none;
   }
 }
+
+.button-reply{
```
rowanseymour: Can we see some screenshots of the simulator changes

temba/channels/tests.py
teehamaral: Yes, I'll change the tests after reviews :)

static/less/simulator.less
teehamaral: Yeah, sure.

screen shot 2017-08-25 at 13 34 28

teehamaral:
Hello, guys. Can you review, please? The code coverage is missing some lines, but I'll do that after the reviews.

karma/flows/test_controllers.coffee
```diff
@@ -765,3 +765,26 @@ describe 'Controllers:', ->
...
           expect(scope.isRuleComplete(rule_test['rule'])).toBe(rule_test['complete'])
+
+          
+     it 'should generate json quick replies to send', ->
+      loadFavoritesFlow()
+
+      actionset = flowService.flow.action_sets[0]
+      action = actionset.actions[0]
+    
+      json_quick_reply = 'base':[{'title': 'Yes', 'payload': 'Yes'}]
```
ericnewcomer: Shouldn't these be the same, ie, no payload?

karma/flows/test_controllers.coffee
```diff
@@ -765,3 +765,26 @@ describe 'Controllers:', ->
...
+
+      editAction actionset, action, (scope) ->
+        scope.actions_buttons_reply = []
+        scope.actions_quick_reply = []
+        scope.action.quick_replies = {}
+        scope.addNewQuickReply()
+        scope.actions_quick_reply[0]['payload'] = 'Yes'
+        scope.actions_quick_reply[0]['title'] = 'Yes'
```
ericnewcomer: Same as above, payload should be removed.

static/coffee/flows/controllers.coffee
```diff
@@ -726,13 +727,28 @@ app.controller 'FlowController', [ '$scope', '$rootScope', '$timeout', '$log', '
 
         fromText = action.msg[Flow.flow.base_language]
 
+        try
+          fromQuickReply = action.quick_replies[Flow.flow.base_language]
```
ericnewcomer: would rather have conditionals instead of using exceptions for flow control

static/coffee/flows/controllers.coffee
```diff
@@ -726,13 +727,28 @@ app.controller 'FlowController', [ '$scope', '$rootScope', '$timeout', '$log', '
...
+          toQuickReply = action.quick_replies[Flow.language.iso_code]
+
+          if typeof toQuickReply == "undefined" && (fromQuickReply != [] && fromQuickReply?)
```
ericnewcomer: This has code smell, why not just inspect if it's there first?

static/coffee/flows/controllers.coffee
```diff
@@ -467,7 +467,7 @@ app.controller 'FlowController', [ '$scope', '$rootScope', '$timeout', '$log', '
         msg: msg
       ]
 
-    @clickAction(actionset, actionset.actions[0])
+    @clickAction(actionset, actionset.actions[0], startNewNode=true)
```
ericnewcomer: Confused on the need for this new variable, can you elaborate?

static/coffee/flows/controllers.coffee
```diff
@@ -655,7 +655,7 @@ app.controller 'FlowController', [ '$scope', '$rootScope', '$timeout', '$log', '
           DragHelper.showSendReply($('#' + category.sources[0] + ' .source'))
         ,0
 
-  $scope.addAction = (actionset) ->
+  $scope.addAction = (actionset, innerAction=null) ->
```
ericnewcomer: Not sure what innerAction could mean for addAction either. Can you help me understand why this is needed?

teehamaral: Maybe the nomenclature has caused confusion, `innerAction` could be a conditional like `isNewStepAction` or something else (when the plus button is tapped on flow step). I needed this argument to control when it needs to restart the quick replies variables in `NodeEditorController`.

teehamaral:
I have added some commits about @ericnewcomer's comments, can you review again, please?

media/test_flows/flow_quick_replies.json
```diff
@@ -0,0 +1,90 @@
...
+          "msg": {
+            "eng": "Hello, choose\u200b an option, please."
+          },
+          "type": "reply",
+          "quick_replies": {
+            "por": [
+              {
+                "title": "Op\u00e7\u00e3o 1"
+              },
+              {}
```
rowanseymour: eh?

media/test_flows/flow_quick_replies.json
```diff
@@ -0,0 +1,90 @@
...
+      "actions": [
+        {
+          "uuid": "d18d5912-eccd-4978-95aa-26f814d28b75",
+          "webhook_headers": [],
+          "action": "GET",
+          "msg": {
+            "eng": "Thank you!"
+          },
+          "type": "reply",
+          "quick_replies": {}
```
rowanseymour: "quick_replies" should be omitted from flow serialization if it is empty

static/coffee/flows/controllers.coffee
```diff
@@ -1805,6 +1843,27 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
+        $scope.actions_quick_reply.push({title:''})
+      else
+        for lang of $scope.action.quick_replies
+          $scope.action.quick_replies[lang].push({title:''})
+
+  $scope.removeElementArrayQuickReply = (a, index) ->
+    for lang of $scope.action.quick_replies
+      $scope.action.quick_replies[lang].splice(index, 1)
+
+    if Object.getOwnPropertyNames($scope.action.quick_replies).length == 0 || $scope.action.quick_replies.length == 0
```
rowanseymour: Any reason this can't be just `Object.keys` like above?

static/coffee/flows/controllers.coffee
```diff
@@ -1805,6 +1843,27 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
       $scope.addNewActionWebhookHeader()
 
+  $scope.addNewQuickReply = ->
+    if $scope.actions_quick_reply.length < 11
+      $scope.container_operation_visible = false
+      if Object.keys($scope.action.quick_replies).length < 1
+        $scope.actions_quick_reply.push({title:''})
+      else
+        for lang of $scope.action.quick_replies
+          $scope.action.quick_replies[lang].push({title:''})
```
rowanseymour: Do we need to push empty translations here? Seems better to only add actual translations to the flow definition

static/coffee/flows/controllers.coffee
```diff
@@ -1851,14 +1910,22 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
+    $scope.container_operation_visible = true
 
   # Saving a reply message in the flow
   $scope.saveMessage = (message, type='reply') ->
-
     if typeof($scope.action.msg) != "object"
       $scope.action.msg = {}
     $scope.action.msg[$scope.base_language] = message
 
+    if Object.prototype.toString.call($scope.action.quick_replies) != '[object Object]'
```
rowanseymour: there's got to be a better way of doing this? this seems hacky

static/coffee/flows/controllers.coffee
```diff
@@ -2118,9 +2185,10 @@ TerminalWarningController = ($scope, $modalInstance, $log, actionset, flowContro
     $modalInstance.dismiss "cancel"
 
 
-AttachmentViewerController = ($scope, $modalInstance, action, type) ->
+AttachmentViewerController = ($scope, $modalInstance, action, type, Flow) ->
```
rowanseymour: We don't need a viewer for quick replies - don't add it to this controller

temba/api/v2/tests.py
```diff
@@ -1853,6 +1853,7 @@ def assertMsgEqual(self, msg_json, msg, msg_type, msg_status, msg_visibility):
             'archived': msg.visibility == 'A',
             'visibility': msg_visibility,
             'text': msg.text,
+            'metadata': msg.metadata if msg.metadata else [],
```
rowanseymour: `else {},`

temba/contacts/models.py
```diff
@@ -1852,7 +1852,7 @@ def raw_tel(self):
         if tel:
             return tel.path
 
-    def send(self, text, user, trigger_send=True, response_to=None, message_context=None, connection=None,
+    def send(self, text, user, trigger_send=True, response_to=None, message_context=None, connection=None, metadata=None,
```
rowanseymour: `metadata` shouldn't be a parameter here - a method like this should take things which end up in metadata, so param here should be `quick_replies`

temba/tests.py
```diff
@@ -121,6 +121,9 @@ def setUpClass(cls):
         with open('%s/test_flows/color.json' % settings.MEDIA_ROOT, 'r') as f:
             cls.COLOR_FLOW_DEFINITION = json.load(f)
 
+        with open('%s/test_flows/flow_quick_replies.json' % settings.MEDIA_ROOT, 'r') as fqr:
+            cls.QUICK_REPLY_FLOW_DEFINITION = json.load(fqr)
```
rowanseymour: This doesn't need to be a class var of `TembaTest`. Color flow is a special case because we use it so often in legacy tests. Just use get_flow to load flow_quick_replies.json as and when you need it.

temba/utils/twitter.py
```diff
@@ -25,7 +25,7 @@ def __init__(self, *args, **kwargs):
         self.events = []
 
     @classmethod
-    def from_channel(cls, channel):
+    def from_channel(cls, channel, headers=None):
```
rowanseymour: Doesn't seem headers is used anywhere?

temba/utils/twitter.py
```diff
@@ -37,15 +37,19 @@ def from_channel(cls, channel):
...
+        return TembaTwython(api_key, api_secret, access_token, access_token_secret, client_args=headers)
 
     def _request(self, url, method='GET', params=None, api_call=None):
         """Internal request method"""
         method = method.lower()
         params = params or {}
 
         func = getattr(self.client, method)
-        params, files = _transparent_params(params)
+        if params.get("event", None):   # TODO: CHANGE THIS LINE BEFORE SEND
```
rowanseymour: What are these TODOs for?

rowanseymour:
Some additional comments on the UI:

* The "Text" label should be "Quick Replies"
* Link to add a new reply should be just "Add reply"
* There should be padding above "Add reply" (see headers on Webhook ruleset for comparison)
* Text box width should be closer to the reply length limit - which we found was 20 for Telegram (check the other services for the lowest!). Maybe boxes should stack horizontally?

screen shot 2017-09-21 at 16 23 44

temba/channels/types/firebase/type.py
```diff
@@ -32,6 +32,16 @@ class FirebaseCloudMessagingType(ChannelType):
     attachment_support = False
     free_sending = True
 
+    @classmethod
+    def get_quick_replies(cls, metadata, post_body):
```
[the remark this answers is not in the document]
teehamaral: The data can be handled on app mobile side. It allows the developer to get this data and implement the buttons how he wants.

temba/channels/types/telegram/type.py
```diff
@@ -46,11 +47,30 @@ def deactivate(self, channel):
...
+        quick_replies = metadata.get('quick_replies', None)
+        replies = []
+
+        if quick_replies:
+            for reply in quick_replies:
+                replies.append([dict(text=reply.get('title'))])
+
+            keyboard_json = dict(resize_keyboard=True, one_time_keyboard=True)
+            keyboard_json['keyboard'] = replies
+            post_body['reply_markup'] = json.dumps(keyboard_json)
```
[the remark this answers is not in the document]
teehamaral: It's being done on backend side because is allowed to use flow expressions and doing that could cut or not complete the flow expression tag. Would you have any idea how do we can implement that limiting the characters but keep the flow expressions complete?

temba/channels/types/twitter/type.py
```diff
@@ -42,6 +43,36 @@ def deactivate(self, channel):
         # tell Mage to deactivate this channel
         notify_mage_task.delay(channel.uuid, MageStreamAction.deactivate.name)
 
+    @classmethod
+    def get_quick_replies(cls, metadata, post_body):
```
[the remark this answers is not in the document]
teehamaral: So, Is it would be `get_payload` instead?

temba/channels/types/telegram/type.py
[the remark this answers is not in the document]
rowanseymour: Oh good point. I guess we just have to let those be truncated by the channels

temba/channels/types/twitter/type.py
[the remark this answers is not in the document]
rowanseymour: I think you just need one method on `Msg` that returns a list of quick replies, and you call that from each channel type's send method.

teehamaral:
Hey guys, could you review, please?

rowanseymour:
@teehamaral hey we just tried to run this locally and it errors as below. Please fix that and we'll take another look.

```
Traceback (most recent call last):
  File "/Users/rowan/Nyaruka/rapidpro/temba/msgs/models.py", line 885, in process_message
    handled = handler.handle(msg)
  File "/Users/rowan/Nyaruka/rapidpro/temba/flows/handlers.py", line 13, in handle
    (handled, msgs) = Flow.find_and_handle(msg)
  File "/Users/rowan/Nyaruka/rapidpro/temba/flows/models.py", line 600, in find_and_handle
    continue_parent=continue_parent)
  File "/Users/rowan/Nyaruka/rapidpro/temba/flows/models.py", line 676, in handle_destination
    result = Flow.handle_actionset(destination, step, run, msg, started_flows, is_test_contact)
  File "/Users/rowan/Nyaruka/rapidpro/temba/flows/models.py", line 729, in handle_actionset
    msgs = actionset.execute_actions(run, msg, started_flows)
  File "/Users/rowan/Nyaruka/rapidpro/temba/flows/models.py", line 3562, in execute_actions
    msgs += action.execute(run, context, self.uuid, msg)
  File "/Users/rowan/Nyaruka/rapidpro/temba/flows/models.py", line 5154, in execute
    quick_replies = ReplyAction.get_translated_metadata(self.quick_replies, run) if self.quick_replies else []
  File "/Users/rowan/Nyaruka/rapidpro/temba/flows/models.py", line 5137, in get_translated_metadata
    for item in metadata.get('quick_replies'):
AttributeError: 'list' object has no attribute 'get'
```

teehamaral:
It shouldn't be happening, I'll check soon.

teehamaral:
Hey @rowanseymour, could you test and review again, please?

rowanseymour:
@teehamaral simulator looks like this for me:

screen shot 2017-10-20 at 13 51 04

teehamaral:
@rowanseymour oh, I'll fix the simulator to work with the quick replies code refactored. Did it work for you with supported channels like Facebook, Telegram?

teehamaral:
@rowanseymour I have fixed the simulator error and added more tests

teehamaral:
@rowanseymour any thoughts? 🙂

rowanseymour:
hey @teehamaral am getting to the code but minor styling comment - can we get rid of the highlighting on the simulator buttons when they're pressed, e.g.

screen shot 2017-10-27 at 16 24 22

temba/api/v2/serializers.py
```diff
@@ -846,6 +847,9 @@ def get_status(self, obj):
     def get_attachments(self, obj):
         return [a.as_json() for a in obj.get_attachments()]
 
+    def get_metadata(self, obj):
```
rowanseymour: Do we need to expose this for now? Seems like metadata schema might change over time

temba/channels/types/facebook/type.py
```diff
@@ -51,10 +51,26 @@ def deactivate_trigger(self, trigger):
...
+        if data:
+            replies = [dict(title=item, payload=item, content_type='text') for item in data]
+
+        return replies
+
     def send(self, channel, msg, text):
         # build our payload
         payload = {'message': {'text': text}}
 
+        metadata = msg.metadata if hasattr(msg, 'metadata') else None
```
rowanseymour: Seems like this could be simpler, e.g.

```python
metadata = json.loads(msg.metadata) if hasattr(msg, 'metadata') else {}
quick_replies = metadata.get('quick_replies', [])
formatted_replies = [dict(title=item, payload=item, content_type='text') for item in quick_replies]
```

But really we should put the metadata into redis as a dict, and pull it out as a dict, so individual channels don't have to parse the JSON each time

temba/channels/types/firebase/type.py
```diff
@@ -31,6 +31,13 @@ class FirebaseCloudMessagingType(ChannelType):
     max_length = 10000
     attachment_support = False
     free_sending = True
+    quick_reply_text_size = 36
+
+    @classmethod
+    def format_quick_replies(cls, quick_replies):
```
rowanseymour: as above for the facebook channel - it seems odd that the arg here is called quick_replies when really it's a serialised version of Msg.metadata

temba/msgs/models.py
```diff
@@ -1421,9 +1459,25 @@ def substitute_variables(cls, text, context, contact=None, org=None, url_encode=
...
+                        channel_type = Channel.get_type_from_code(channel.channel_type)
+                        quick_replies[counter] = text[:channel_type.quick_reply_text_size]
+
+        return json.dumps(metadata)
+
+    @classmethod
     def create_outgoing(cls, org, user, recipient, text, broadcast=None, channel=None, high_priority=False,
                         created_on=None, response_to=None, message_context=None, status=PENDING, insert_object=True,
-                        attachments=None, topup_id=None, msg_type=INBOX, connection=None):
+                        attachments=None, topup_id=None, msg_type=INBOX, connection=None, metadata=None):
```
rowanseymour: this arg should be called quick_replies rather than metadata since quick_replies is just one item we put we put in metadata

temba/msgs/models.py
```diff
@@ -726,6 +753,8 @@ class Msg(models.Model):
     connection = models.ForeignKey('channels.ChannelSession', null=True,
                                    help_text=_("The session this message was a part of if any"))
 
+    metadata = models.TextField(null=True, help_text=_("The metadata for any type msgs"))
```
rowanseymour: "The metadata for this msg"

temba/flows/models.py
```diff
@@ -5152,6 +5178,10 @@ def execute(self, run, context, actionset_uuid, msg, offline_on=None):
             if self.msg:
                 text = run.flow.get_localized_text(self.msg, run.contact)
 
+            quick_replies = ReplyAction.get_translated_metadata(self.quick_replies, run) if self.quick_replies else []
+
+            metadata = json.dumps(dict(quick_replies=quick_replies))
```
rowanseymour: We should serialise to JSON only when saving message object

temba/msgs/models.py
```diff
@@ -1045,12 +1074,17 @@ def mark_sent(cls, r, msg, status, external_id=None):
...
-                    model="msg")
+        msg_json = dict(direction=self.direction,
+                        text=self.text,
+                        id=self.id,
+                        attachments=self.attachments,
+                        created_on=self.created_on.strftime('%x %X'),
+                        model="msg")
+
+        if self.metadata:
+            msg_json['metadata'] = self.metadata
```
rowanseymour: I think we load this and include it as a nested dict

rowanseymour: Yeah that's probably simpler - so if it isn't set in a message, it should go into the payload as `{}`

teehamaral:
@rowanseymour I have updated the code, could you review again, please? Thanks.

johncordeiro:
hey @nicpottier, is it ready to merge?

ericnewcomer:
We had holidays here at the end of last week, we'll have another look today.

static/coffee/flows/controllers.coffee
```diff
@@ -1836,6 +1847,20 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
+
+  if $scope.options.dragSource?
+    startNodeConfig()
+  else
+    if $scope.action.quick_replies? and $scope.action.quick_replies != undefined and $scope.action.quick_replies.length > 0
+      $scope.quickReplies = $scope.action.quick_replies
+      $scope.showQuickReplyButton = false
+    else
+      startNodeConfig()
+
```
ericnewcomer: This is really confusing. Maybe just initialize it first and set it if appropriate.

```coffeescript
$scope.quickReplies = []
$scope.action.quick_replies = []
$scope.showQuickReplyButton = true

if $scope.action.quick_replies? and $scope.action.quick_replies.length > 0
    $scope.quickReplies = $scope.action.quick_replies
    $scope.showQuickReplyButton = false

```

static/coffee/simulator.coffee
```diff
@@ -42,6 +42,17 @@ window.updateSimulator = (data) ->
     media_type = null
     media_viewer_elt = null
 
+    quick_replies = null
+
+    if msg.metadata and Object.keys(msg.metadata).length > 1
```
ericnewcomer: This should be checking for the quick replies key, not on the total length of metadata which can contain who knows what.

static/coffee/simulator.coffee
```diff
@@ -42,6 +42,17 @@ window.updateSimulator = (data) ->
...
+      params = JSON.parse(msg.metadata)
+
+      if params.quick_replies? and params.quick_replies[0]?
```
ericnewcomer: Too late for the undefined operator here if the length is zero.

static/fonts/temba/style.css
```diff
@@ -789,4 +789,9 @@
 .icon-chatbase:before {
   content: "\e902";
 }
-
+.icon-button-quick-reply:before {
+  content: "\e903";
+}
+.icon-button-response:before {
+  content: "\e907";
+}
```
ericnewcomer: There should be a selection.json file for the font included as well.

static/less/flow_editor.less
```diff
@@ -2214,4 +2249,88 @@ div.modal[role="dialog"] {
...
+    padding: 14px 6px;
+    margin: 5px 0 0 0;
+    &.ng-invalid-maxlength {
+      border-color: #e9322d;
+      -webkit-box-shadow: 0 0 6px #f8b9b7;
+      -moz-box-shadow: 0 0 6px #f8b9b7;
+      box-shadow: 0 0 6px #f8b9b7;
+    }
+  }
+  .quick_reply-title {
```
ericnewcomer: Mixed hyphen and underscore

static/less/flow_editor.less
```diff
@@ -2214,4 +2249,88 @@ div.modal[role="dialog"] {
...
+    display: inline-block;
+    width: 95%;
+    margin: 0 1% 0 0;
```
ericnewcomer: The alignment looks a bit off in the UI. How about: `margin: 5px 0 0 0;`

temba/api/v2/views.py
```diff
@@ -2208,10 +2209,11 @@ class MessagesEndpoint(ListAPIMixin, BaseAPIView):
                 "status": "wired",
                 "visibility": "visible",
                 "text": "How are you?",
-                "media": "wav:http://domain.com/recording.wav"
+                "media": "wav:http://domain.com/recording.wav",
+                "metadata": {},
```
ericnewcomer: remove references to metadata from api

temba/channels/tests.py
```diff
@@ -7064,6 +7064,30 @@ def test_send_media(self):
             self.assertEqual(mock.call_args[0][1]['caption'], "Test message")
             self.assertEqual(mock.call_args[0][1]['chat_id'], "1234")
 
+    def test_send_quick_replies(self):
+        metadata = dict(quick_replies=['Yes', 'No'])
+        joe = self.create_contact("Ernie", urn='telegram:1234')
+        msg = joe.send("Test message", self.admin, trigger_send=False, quick_replies=metadata)[0]
+
+        settings.SEND_MESSAGES = True
```
ericnewcomer: Correct pattern for this going forward should instead be:

`@override_settings(SEND_MESSAGES=True)` as a decorator for your method, then ditch the `finally`, etc

temba/channels/tests.py
```diff
@@ -7508,6 +7532,73 @@ def test_send(self):
         finally:
             settings.SEND_MESSAGES = False
 
+    def test_send_quick_replies(self):
+        metadata = dict(quick_replies=['Yes', 'No'])
+        msg = self.joe.send("Hello, world!", self.admin, trigger_send=False, quick_replies=metadata)[0]
+
+        try:
+            settings.SEND_MESSAGES = True
```
ericnewcomer: as above, you copied a bad example ;)

temba/channels/tests.py
```diff
@@ -9458,6 +9584,55 @@ def test_send_media(self):
             self.assertFalse(ChannelLog.objects.filter(description__icontains="local variable 'response' "
                                                                               "referenced before assignment"))
 
+    def test_send_quick_replies(self):
+        joe = self.create_contact("Joe", urn="facebook:1234")
+        metadata = dict(quick_replies=['Yes', 'No'])
+        msg = joe.send("Facebook Msg", self.admin, trigger_send=False, quick_replies=metadata)[0]
+
+        with self.settings(SEND_MESSAGES=True):
```
ericnewcomer: Same thing here, let's be consistent in using `@override_settings` for this instead

temba/channels/tests.py
```diff
@@ -10612,6 +10787,55 @@ def test_send_media(self):
             self.assertEqual(msg.external_id, "4987381194038857789")
             self.clear_cache()
 
+    def test_send_quick_replies(self):
+        metadata = dict(quick_replies=['Yes', 'No'])
+        joe = self.create_contact("Joe", urn="viber:FXLP/JstS7kDuoiUGihkgA==")
+        msg = joe.send("Hello, world!", self.admin, trigger_send=False, quick_replies=metadata)[0]
+
+        settings.SEND_MESSAGES = True
```
ericnewcomer: `@override_settings`

temba/channels/tests.py
```diff
@@ -10782,6 +11006,52 @@ def test_send_media(self):
 
                 self.clear_cache()
 
+    def test_send_quick_replies(self):
+        metadata = dict(quick_replies=['Yes', 'No'])
+        joe = self.create_contact("Joe", urn="fcm:12345abcde", auth="123456abcdef")
+        msg = joe.send("Hello, world!", self.admin, trigger_send=False, quick_replies=metadata)[0]
+
+        with self.settings(SEND_MESSAGES=True):
```
ericnewcomer: `@override_settings`

temba/channels/types/firebase/type.py
```diff
@@ -56,6 +57,12 @@ def send(self, channel, msg, text):
             }
             data['content_available'] = True
 
+        metadata = msg.metadata if hasattr(msg, 'metadata') else {}
+        quick_replies = metadata.get('quick_replies', [])
+
+        if quick_replies:
+            data['data']['metadata'] = quick_replies
+
```
ericnewcomer: Few questions here:

1) Why is this in firebase at all? Not sure I understand how that would surface on the Android Channel which is what firebase is used for.

2) Why is this stuffing quick_replies in something called metadata? Is there a spec somewhere you can share? Feels weird that we get quick_replies out of something we call metadata and then assign it to somebody else's metadata.

temba/channels/types/twitter/type.py
```diff
@@ -56,10 +57,37 @@ def send(self, channel, msg, text):
...
+                                message_data=dict(
+                                    text=text,
+                                    quick_reply=dict(
+                                        type='options',
+                                        options=formatted_replies
+                                    )
+                                )
+                            )
+                        )
+                    )
```
ericnewcomer: This looks a bit nutty. If you want to indent this, just use `{}`

temba/contacts/models.py
```diff
@@ -1893,7 +1893,7 @@ def send(self, text, user, trigger_send=True, response_to=None, expressions_cont
                 msg = Msg.create_outgoing(self.org, user, recipient, text,
                                           response_to=response_to, expressions_context=expressions_context, connection=connection,
                                           attachments=attachments, msg_type=msg_type or INBOX, status=status,
-                                          created_on=created_on, high_priority=high_priority)
+                                          metadata=quick_replies, created_on=created_on, high_priority=high_priority)
```
ericnewcomer: create_outgoing should have `quick_replies` as a parameter. We shouldn't ever see `metadata=quick_replies`

temba/flows/models.py
```diff
@@ -1644,6 +1644,13 @@ def start_call_flow(self, all_contact_ids, start_msg=None, extra=None, flow_star
 
         return runs
 
+    @classmethod
+    def get_send_action_metadata(cls, send_action):
+        if send_action.quick_replies:
+            return dict(quick_replies=send_action.quick_replies)
+        else:
+            return None
+
```
ericnewcomer: Do we really need this method? This is weird.

temba/flows/models.py
```diff
@@ -1665,10 +1672,13 @@ def start_msg_flow(self, all_contact_ids, started_flows=None, start_msg=None, ex
...
                 if (send_action.msg and send_action.msg.get(self.base_language)) or (send_action.media and send_action.media.get(self.base_language)):
 
+                    metadata = Flow.get_send_action_metadata(send_action)
+
                     broadcast = Broadcast.create(self.org, self.created_by, send_action.msg, [],
                                                  media=send_action.media,
                                                  base_language=self.base_language,
-                                                 send_all=send_action.send_all)
+                                                 send_all=send_action.send_all,
+                                                 metadata=metadata)
```
ericnewcomer: Lets pass quick_replies down instead of all metadata, would rather this be explicit.

temba/flows/models.py
```diff
@@ -1759,7 +1769,7 @@ def start_msg_flow_batch(self, batch_contact_ids, broadcasts, started_flows, sta
                 # create the sms messages
                 created_on = timezone.now()
                 broadcast.send(expressions_context=expressions_context_base, trigger_send=False,
-                               response_to=start_msg, status=INITIALIZING, msg_type=FLOW,
+                               response_to=start_msg, status=INITIALIZING, msg_type=FLOW, metadata=broadcast.metadata,
```
ericnewcomer: as above

temba/flows/models.py
```diff
@@ -5297,10 +5309,24 @@ def from_json(cls, org, json_obj):
...
+    def get_translated_metadata(metadata, run):
+        """
+        Gets the appropriate metadata translation for the given contact
+        """
+        language_metadata = []
+        for item in metadata:
+            quick_reply = item.get(run.contact.language) \
+                if run.contact.language and run.contact.language in item else item.get(run.flow.base_language)
+            language_metadata.append(quick_reply)
+
```
ericnewcomer: rename fail? This should be `get_translated_quick_replies`, with properly named variables therein.

temba/flows/models.py
```diff
@@ -5312,6 +5338,9 @@ def execute(self, run, context, actionset_uuid, msg, offline_on=None):
             if self.msg:
                 text = run.flow.get_localized_text(self.msg, run.contact)
 
+            quick_replies = ReplyAction.get_translated_metadata(self.quick_replies, run) if self.quick_replies else []
+            metadata = dict(quick_replies=quick_replies)
+
```
ericnewcomer: Since we'll just be passing quick_replies down below, don't need to construct metadata here.

temba/flows/tests.py
```diff
@@ -2693,6 +2694,33 @@ def test_flow_start_with_start_msg(self):
...
+        self.assertEqual(actionset_step.messages.all().count(), 1)
+
+        runs = flow.start_msg_flow([self.contact3.id, self.contact4.id])
+        self.assertEqual(2, len(runs))
+
+        contact_test = self.create_contact('Teeh', '+250788123457', language='por', is_test=True)
+        flow.start([], [contact_test])
+        self.assertTrue(FlowRun.objects.filter(contact=contact_test))
+
+        runs = flow.start_msg_flow([contact_test.id])
```
ericnewcomer: Why not just `flow.start()`

temba/msgs/models.py
```diff
@@ -378,15 +381,34 @@ def get_translated_text(self, contact, org=None):
...
+        Gets the appropriate metadata translation for the given contact
+        """
+        preferred_languages = self.get_preferred_languages(contact, org)
+        language_metadata = []
+        for item in metadata.get('quick_replies'):
+            current_language = preferred_languages[0] if preferred_languages else None
+            quick_reply = item.get(current_language) if current_language in item else item.get(self.base_language)
+            language_metadata.append(quick_reply)
+
+        return language_metadata
```
ericnewcomer: Same as before, this should probably be `get_translated_quick_replies`

temba/msgs/models.py
```diff
@@ -483,7 +508,8 @@ def send(self, trigger_send=True, expressions_context=None, response_to=None, st
                                           high_priority=high_priority,
                                           insert_object=False,
                                           attachments=[media] if media else None,
-                                          created_on=created_on)
+                                          created_on=created_on,
+                                          metadata=_metadata)
```
ericnewcomer: `quick_replies`

temba/msgs/models.py
```diff
@@ -1117,11 +1146,12 @@ def get_last_log(self):
         return sorted_logs[0] if sorted_logs else None
 
     def reply(self, text, user, trigger_send=False, expressions_context=None, connection=None, attachments=None, msg_type=None,
-              send_all=False, created_on=None):
+              send_all=False, created_on=None, metadata=None):
```
ericnewcomer: `quick_replies`

temba/msgs/models.py
```diff
@@ -378,15 +381,34 @@ def get_translated_text(self, contact, org=None):
...
         return Language.get_localized_text(self.text, preferred_languages)
 
+    def get_translated_metadata(self, metadata, contact, org=None):
+        """
+        Gets the appropriate metadata translation for the given contact
+        """
+        preferred_languages = self.get_preferred_languages(contact, org)
+        language_metadata = []
+        for item in metadata.get('quick_replies'):
+            current_language = preferred_languages[0] if preferred_languages else None
```
ericnewcomer: This shouldn't be in the loop.

Also should be using Language.get_localized_text() which is smart about how things backdown when languages are missing.

temba/flows/models.py
```diff
@@ -5297,10 +5309,24 @@ def from_json(cls, org, json_obj):
...
+                    send_all=self.send_all)
+
+    @staticmethod
+    def get_translated_metadata(metadata, run):
+        """
+        Gets the appropriate metadata translation for the given contact
+        """
+        language_metadata = []
+        for item in metadata:
+            quick_reply = item.get(run.contact.language) \
```
ericnewcomer: Should be using `run.flow.get_localized_text()` which is smart about how things backdown when languages are missing.

temba/msgs/models.py
```diff
@@ -1400,9 +1433,24 @@ def evaluate_template(cls, text, context, org=None, url_encode=False, partial_va
...
+        if quick_replies:
+            for counter, reply in enumerate(quick_replies):
+                (text, errors) = Msg.evaluate_template(reply, expressions_context, org=org)
+                if text:
+                    if contact.is_test:
+                        quick_replies[counter] = text
+                    else:
+                        channel_type = Channel.get_type_from_code(channel.channel_type)
+                        quick_replies[counter] = text[:channel_type.quick_reply_text_size]
+
```
ericnewcomer: Like to avoid in-place changes, instead create a new list and return that. Also, looks like this classmethod is used only once. Is that intentional?

temba/msgs/models.py
```diff
@@ -1400,9 +1433,24 @@ def evaluate_template(cls, text, context, org=None, url_encode=False, partial_va
...
+        return json.dumps(metadata)
+
+    @classmethod
     def create_outgoing(cls, org, user, recipient, text, broadcast=None, channel=None, high_priority=False,
                         created_on=None, response_to=None, expressions_context=None, status=PENDING, insert_object=True,
-                        attachments=None, topup_id=None, msg_type=INBOX, connection=None):
+                        attachments=None, topup_id=None, msg_type=INBOX, connection=None, metadata=None):
```
ericnewcomer: `quick_replies`

static/coffee/flows/controllers.coffee
teehamaral: Doing that the conditional always would be `False` because `$scope.action.quick_replies` is initialized as empty

temba/channels/types/firebase/type.py
teehamaral: I agree about the "metadata" nomenclature, it could be `quick_replies`, right? But about the data of quick replies, this can be handled on the mobile app side, that allows the developer implement the quick replies buttons like other platforms.

temba/flows/models.py
teehamaral: If we work with all metadata, I think that in the future when we want to add one more key to metadata it would be easier, wouldn't be necessary change all send methods to do that for example.

rowanseymour: Think of metadata as being the storage mechanism for some of these different message options. The calling code shouldn't have to worry about how these things are stored - it just sends the message options it wants, and it's up to `Msg` to decide what is a field, what is embedded in metadata etc.

teehamaral: Let's see if I understood, it should be applied just to Broadcast.send() method, the create method stores all metadata but send method only pass quick_replies, right?

rowanseymour: Internally in Msg and Broadcast we can pass around metadata, but calling code should be dealing with higher-level params like quick_replies

temba/flows/models.py
teehamaral: I think that here we could pass `metadata` because internally it's being handled (https://github.com/Ilhasoft/rapidpro/blob/133993c1b482a53e70d9358d69d094f207042322/temba/msgs/models.py#L471)

rowanseymour: Why is metadata needed here - aren't the the broadcasts are being created with the quick_replies at https://github.com/rapidpro/rapidpro/pull/600/files/133993c1b482a53e70d9358d69d094f207042322#diff-ac952ec358ed4e019095b25fcac469adR1668 ?

teehamaral:
@rowanseymour and @ericnewcomer I have updated the code, could you review, please? Thanks.

teehamaral:
Hello guys, any thoughts?

static/coffee/flows/controllers.coffee
```diff
@@ -1873,6 +1898,24 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
 
+  $scope.addNewQuickReply = ->
+    $scope.showQuickReplyButton = false
+    if $scope.quickReplies.length < 11
+      addQuickReply = {}
+      addQuickReply[$scope.base_language] = ''
+      $scope.quickReplies.push(addQuickReply)
+
+  $scope.removeQuickReply = (a, index) ->
+    a.splice(index, 1)
```
rowanseymour: Can we give this a real variable name instead of `a`

static/coffee/flows/controllers.coffee
```diff
@@ -1873,6 +1898,24 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
+
+    if Object.keys($scope.action.quick_replies).length == 0
+      $scope.quickReplies.splice(index, 1)
```
rowanseymour: do we even need `a` since you're also modifying `$scope.quickReplies` directly? Aren't they the same thing?

static/less/flow_editor.less
```diff
@@ -2214,4 +2249,88 @@ div.modal[role="dialog"] {
...
     }
   }
+}
+
+.modal {
+  .cursor-pointer {
+    cursor: pointer;
+  }
+  .cursor-pointer:hover {
+    opacity: 0.7;
```
rowanseymour: If you want a hover style for the quick reply button, then you should apply it to the attachment button to for consistency

temba/channels/tests.py
```diff
@@ -7121,6 +7144,73 @@ def test_send(self):
...
+                self.assertTrue(msg.sent_on)
+                self.assertEqual(msg.external_id, "000000000000000000")
+                self.assertEqual(json.loads(msg.metadata), dict(quick_replies=quick_replies))
+                data_args = json.loads(mock.call_args[1]['data'])
+                message_data = data_args['event']['message_create']['message_data']
+                self.assertEqual(message_data['quick_reply']['options'][0]['label'], 'Yes')
+                self.assertEqual(message_data['quick_reply']['options'][1]['label'], 'No')
+                self.clear_cache()
+
+        finally:
```
rowanseymour: You can get rid of this try/finally since you have @override_settings now

temba/msgs/models.py
```diff
@@ -1258,6 +1287,10 @@ def as_task_json(self):
             data.update(dict(chatbase_api_key=chatbase_api_key, chatbase_version=chatbase_version,
                              is_org_connected_to_chatbase=True))
 
+        if self.metadata:
+            metadata = json.loads(self.metadata)
+            data['quick_replies'] = metadata.get('quick_replies', None)
```
rowanseymour: So I know we told you guys you should be passing around params like `quick_replies` in `Msg` methods - but the Msg struct passed to channels should mirror Msg database objects. So this should be adding `metadata` to `data`.

temba/channels/types/facebook/type.py
```diff
@@ -55,6 +55,12 @@ def send(self, channel, msg, text):
         # build our payload
         payload = {'message': {'text': text}}
 
+        quick_replies = msg.quick_replies if hasattr(msg, 'quick_replies') else {}
```
rowanseymour: See comment on `Msg.as_task_json()`

temba/channels/types/firebase/type.py
```diff
@@ -56,6 +57,11 @@ def send(self, channel, msg, text):
             }
             data['content_available'] = True
 
+        quick_replies = msg.quick_replies if hasattr(msg, 'quick_replies') else {}
```
rowanseymour: See comment on `Msg.as_task_json()`

temba/channels/types/telegram/type.py
```diff
@@ -50,6 +51,13 @@ def send(self, channel, msg, text):
         send_url = 'https://api.telegram.org/bot%s/sendMessage' % auth_token
         post_body = {'chat_id': msg.urn_path, 'text': text}
 
+        quick_replies = msg.quick_replies if hasattr(msg, 'quick_replies') else {}
```
rowanseymour: See comment on `Msg.as_task_json()`

temba/channels/types/twitter/type.py
```diff
@@ -56,10 +57,31 @@ def send(self, channel, msg, text):
             # this is a legacy URN (no display), the path is our screen name
             if scheme == TWITTER_SCHEME:
                 dm = twitter.send_direct_message(screen_name=path, text=text)
+                external_id = dm['id']
 
             # this is a new twitterid URN, our path is our user id
             else:
-                dm = twitter.send_direct_message(user_id=path, text=text)
+                quick_replies = msg.quick_replies if hasattr(msg, 'quick_replies') else {}
```
rowanseymour: See comment on `Msg.as_task_json()`

temba/channels/types/viber_public/type.py
```diff
@@ -58,6 +59,13 @@ def send(self, channel, msg, text):
             'tracking_data': msg.id
         }
 
+        quick_replies = msg.quick_replies if hasattr(msg, 'quick_replies') else {}
```
rowanseymour: See comment on `Msg.as_task_json()`

temba/flows/models.py
```diff
@@ -5528,6 +5545,9 @@ def execute(self, run, context, actionset_uuid, msg, offline_on=None):
             if self.msg:
                 text = run.flow.get_localized_text(self.msg, run.contact)
 
+            quick_replies = ReplyAction.get_translated_quick_replies(self.quick_replies, run) \
```
rowanseymour: Don't use ternary statements that span multiple lines - these are easier to read as regular if statements

temba/msgs/models.py
```diff
@@ -1032,7 +1059,8 @@ def as_json(self):
                     id=self.id,
                     attachments=self.attachments,
                     created_on=self.created_on.strftime('%x %X'),
-                    model="msg")
+                    model="msg",
+                    metadata=self.metadata or {})
```
rowanseymour: This should be loaded with `json.loads`

temba/msgs/models.py
```diff
@@ -1396,10 +1429,23 @@ def evaluate_template(cls, text, context, org=None, url_encode=False, partial_va
...
+    @classmethod
+    def get_outgoing_metadata(cls, quick_replies, expressions_context, contact, org, channel):
+        for counter, reply in enumerate(quick_replies):
+            (text, errors) = Msg.evaluate_template(reply, expressions_context, org=org)
+            if text:
+                if contact.is_test:
+                    quick_replies[counter] = text
+                else:
+                    channel_type = Channel.get_type_from_code(channel.channel_type)
+                    quick_replies[counter] = text[:channel_type.quick_reply_text_size]
```
rowanseymour: This should happen in each channel handler - which means this method is probably unnecessary. Just call Msg.evaluate_template on each quick_reply in Msg.create_outgoing

templates/flows/flow_editor.haml
```diff
@@ -386,6 +386,7 @@
 
                     .node-attachment.viewer-button{ng-class:'{"icon-videocam": action._media.type == "video", "icon-photo": action._media.type=="image", "icon-arrow-right-8": action._media.type=="audio"}', ng-show:'action._media', ng-click:'clickShowActionMedia(action); $event.stopPropagation();'}
                     .node-attachment.icon-attachment{ng-show:'action._attachURL'}
+                    .node-attachment.viewer-button{ng-class:'{"icon-button-quick-reply": action.quick_replies.length > 0}', ng-show:'action.quick_replies.length > 0'}
```
rowanseymour: css class doesn't need to be conditional here, i.e.

`.node-attachment.viewer-button.icon-button-quick-reply`

templates/partials/translation_modal.haml
```diff
@@ -24,6 +29,16 @@
...
 
+            .quick-reply{ng-show:"translation.fromQuickReplies.length"}
+              .modal-quick-reply-header
+                .quick-reply-title
+                  -trans "Quick Replies"
+
+              .modal-quick-reply{ng-repeat:"response in translation.fromQuickReplies", class:""}
+                .quick-reply-title.center-translate
+                  .translate-from{ng-show:"translation.fromQuickReplies"}
+                    -trans "[[ response[language.from] ]]"
```
rowanseymour: You shouldn't be using `-trans` here - it can't localize an angular expression

teehamaral:
Hey @rowanseymour, I have updated the code, could you check it, please? Thanks.

temba/msgs/models.py
```diff
@@ -1258,6 +1287,10 @@ def as_task_json(self):
             data.update(dict(chatbase_api_key=chatbase_api_key, chatbase_version=chatbase_version,
                              is_org_connected_to_chatbase=True))
 
+        if self.metadata:
+            metadata = json.loads(self.metadata)
+            data['metadata'] = metadata.get('quick_replies', None)
```
rowanseymour: data['metadata'] = json.loads(self.metadata)

temba/msgs/models.py
```diff
@@ -1506,6 +1539,18 @@ def create_outgoing(cls, org, user, recipient, text, broadcast=None, channel=Non
...
 
+        metadata = None
+        if quick_replies:
+            for counter, reply in enumerate(quick_replies):
+                (value, errors) = Msg.evaluate_template(text=reply, context=expressions_context, org=org)
+                if value:
+                    if contact.is_test:
+                        quick_replies[counter] = value
+                    else:
+                        channel_type = Channel.get_type_from_code(channel.channel_type)
```
rowanseymour: As we said previously - this truncation should be happening in the channel handler

static/coffee/flows/controllers.coffee
```diff
@@ -1873,6 +1892,21 @@ NodeEditorController = ($rootScope, $scope, $modalInstance, $timeout, $log, Flow
...
+      addQuickReply = {}
+      addQuickReply[$scope.base_language] = ''
+      $scope.quickReplies.push(addQuickReply)
+
+  $scope.removeQuickReply = (item, index) ->
+    item.splice(index, 1)
+
+    if item.length == 0
+      $scope.showQuickReplyButton = true
+      $scope.quickReplies = []
```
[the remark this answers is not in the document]
teehamaral: Oh, I got it now.

teehamaral:
@rowanseymour I think that your comments were resolved, can you check, please? Thanks.

static/coffee/simulator.coffee
```diff
@@ -42,6 +42,21 @@ window.updateSimulator = (data) ->
...
     media_viewer_elt = null
 
+    quick_replies = null
+
+    metadata = msg.metadata
+    if metadata
+      if typeof(metadata) == "object"
+        metadata = JSON.stringify(metadata)
+
+      params = JSON.parse(metadata)
```
ericnewcomer: why is this stringifying and then parsing here?

teehamaral:
@ericnewcomer @rowanseymour do we have your thumbs up for that? 😁

rowanseymour:
@teehamaral nice work 🍾

teehamaral:
@rowanseymour @ericnewcomer @nicpottier Thanks guys for your help on that, great teamwork! ✌️😉

nicpottier:
Although I'm psyched we finally got this in, @teehamaral I'd like to ask that your team spend a bit more team on code quality before submitting PRs. This PR took much longer than it should have to get to a mergeable state and even with that some bugs snuck in. (see https://github.com/nyaruka/rapidpro/pull/1638 and https://github.com/rapidpro/rapidpro/issues/707)

We shouldn't be the only line of defense on these PRs, only the LAST line of defense. So please have your team do internal reviews as well instead of leaning so heavily on us for ensuring things are built in the right way. You should also be testing the code yourself in production and on your staging servers etc, so that you have confidence that the implementation you are putting into mainline is correct. (as we have done when we merge into RapidPro)

As it stands, we are very hesitant to accept PRs from Ihlasoft for large bodies of work because on every instance we have done so large bugs have been introduced. Smaller changes are easier for us to reason about and make sure are correct, but we'll need to see higher quality in the submissions going forward to feel comfortable with accepting changes of this scope in the future.

PR #602, opened 2017-07-21
Wait for transaction commit before pushing tasks

erikh360:
We were experiencing issues with our USSD service, our sentry was showing these errors:

```
IntegrityError: insert or update on table "channels_channellog" violates foreign key constraint "channels_channellog_msg_id_e40e6612_fk_msgs_msg_id"
DETAIL:  Key (msg_id)=(260040) is not present in table "msgs_msg".
```

After investigating I found the message object would always exist in the DB, but the queue started processing the message before the original transaction was committed. It will now wait for the transaction to commit before placing things on the queue.

We have this fix running in one of our Production environments and the error has disappeared.

### DIFF
--- a/temba/msgs/models.py
+++ b/temba/msgs/models.py
@@ -738,8 +738,6 @@ class Msg(models.Model):
         task_priority = None
         last_contact = None
 
-        tasks_to_push = []
-
         # we send in chunks of 1,000 to help with contention
         for msg_chunk in chunk_list(all_msgs, 1000):
             # create a temporary list of our chunk so we can iterate more than once
@@ -770,7 +768,7 @@ class Msg(models.Model):
                             if task_priority is None:  # pragma: needs cover
                                 task_priority = DEFAULT_PRIORITY
 
-                            tasks_to_push.append((task_msgs, task_priority))
+                            on_transaction_commit(lambda: push_task(task_msgs[0]['org'], MSG_QUEUE, SEND_MSG_TASK, task_msgs, priority=task_priority))
                             task_msgs = []
                             task_priority = None
 
@@ -792,14 +790,7 @@ class Msg(models.Model):
             if task_priority is None:
                 task_priority = DEFAULT_PRIORITY
 
-            tasks_to_push.append((task_msgs, task_priority))
-
-        on_transaction_commit(lambda: cls.push_tasks(tasks_to_push))
-
-    @classmethod
-    def push_tasks(cls, tasks):
-        for msg_task in tasks:
-            push_task(msg_task[0][0]['org'], MSG_QUEUE, SEND_MSG_TASK, msg_task[0], priority=msg_task[1])
+            on_transaction_commit(lambda: push_task(task_msgs[0]['org'], MSG_QUEUE, SEND_MSG_TASK, task_msgs, priority=task_priority))
 
     @classmethod
     def process_message(cls, msg):
--- a/temba/msgs/models.py
+++ b/temba/msgs/models.py
@@ -738,6 +738,8 @@ class Msg(models.Model):
         task_priority = None
         last_contact = None
 
+        tasks_to_push = []
+
         # we send in chunks of 1,000 to help with contention
         for msg_chunk in chunk_list(all_msgs, 1000):
             # create a temporary list of our chunk so we can iterate more than once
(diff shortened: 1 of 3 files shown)


erikh360:
@nicpottier Please review, this is the fix for the discussion we had titled `USSD timing out and/or ending prematurely`, thanks.

temba/settings.py.dev
```diff
@@ -51,6 +51,18 @@ DATABASES = {
...
+        'ENGINE': 'django.contrib.gis.db.backends.postgis',
+        'NAME': 'temba',
+        'USER': 'temba',
+        'PASSWORD': 'temba',
+        'HOST': 'localhost',
+        'PORT': '',
+        'ATOMIC_REQUESTS': True,
+        'CONN_MAX_AGE': 60,
+        'OPTIONS': {
+        }
```
nicpottier: What is this addition for?

temba/msgs/tests.py
```diff
@@ -2103,3 +2107,120 @@ def test_render(self):
         # exception if tag not used correctly
         self.assertRaises(ValueError, self.render_template, '{% load sms %}{% render with bob %}{% endrender %}')
         self.assertRaises(ValueError, self.render_template, '{% load sms %}{% render as %}{% endrender %}')
+
+
+class CeleryTaskTest(TransactionTestCase):
```
nicpottier: I think you could avoid a lot of the boilerplate of setting up an org etc, if you just subclass `TembaTest`

temba/msgs/models.py
```diff
@@ -789,7 +791,15 @@ def send_messages(cls, all_msgs):
...
             if task_priority is None:
                 task_priority = DEFAULT_PRIORITY
-            push_task(task_msgs[0]['org'], MSG_QUEUE, SEND_MSG_TASK, task_msgs, priority=task_priority)
+
+            tasks_to_push.append((task_msgs, task_priority))
+
+        on_transaction_commit(lambda: cls.push_tasks(tasks_to_push))
+
+    @classmethod
+    def push_tasks(cls, tasks):
```
nicpottier: I don't think there's any real need for this, it is basically just recreating what `on_transaction_commit` already does. (it has an internal list, so you can call `on_transaction_commit` more than once)

Just changing all the push_task to `on_transaction_commit` should work no? (being careful about shared variables)

temba/settings.py.dev
erikh360: I use it in the tests inside the `assertInDB` function, this opens a separate DB transaction to make sure the record is actually committed.

temba/msgs/tests.py
erikh360: I had to use the `TransactionTestCase` to get the DB transaction tests working. It's the only way I could get the original code to fail with the tests.

erikh360:
@nicpottier Ready for another look!

temba/msgs/tests.py
nicpottier: Hrmm.. any chance you could use insane Python multiple inheritance to get the best of both? Looking at `TransactionTestCase` I don't think its magic overlaps much with what is provided by TembaTest so I don't think they would walk over each other.

Sorry to put this back in your court but man our tests are already so full of repetition already we are trying not to have that debt grow even more.

erikh360: I'll see what I can do. 🙂

erikh360: @nicpottier Please have a look now. `TembaTest` actually inherits from `TransactionTestCase` a few levels up, so I just had to disable the rollback things.

temba/msgs/tests.py
```diff
@@ -2103,3 +2109,84 @@ def test_render(self):
...
+        try:
+            type(obj).objects.using('default2').get(pk=obj.pk)
+        except obj.DoesNotExist:
+            self.fail(fullmsg)
+
+    def test_reply_task_added(self):
+        send_task_orig = current_app.send_task
+        orig_push_task = models.push_task
+
+        settings.CELERY_ALWAYS_EAGER = False
```
rowanseymour: Use @override_settings decorator on the test method so these automatically get reset

temba/msgs/tests.py
```diff
@@ -2103,3 +2109,84 @@ def test_render(self):
...
+        settings.SEND_MESSAGES = True
+
+        try:
+            def new_send_task(name, args=(), kwargs={}, **opts):  # pragma: needs cover
+                self.handle_push_task(name)
+
+            # monkey patch the regular send_task with our method
+            current_app.send_task = new_send_task
```
rowanseymour: Use `with patch(...):`

erikh360:
@rowanseymour Please have a look again.

rowanseymour:
Thanks, will be good to get this deployed!

Issue #592, opened 2017-07-07
Add support for m.me referrals for Facebook


nicpottier:
Facebook has added support for shortened m.me links to create links to start a chat with a bot:
   https://developers.facebook.com/docs/messenger-platform/referral-params

Users can add a `?ref=params` to the m.me link and Facebook will notify via the webhook what that referral id was.

Right now we support a few different Facebook referrals systems (see https://github.com/nyaruka/rapidpro/blob/master/temba/channels/handlers.py#L2310), but not this one and it seems to be gaining traction so would be good to add.

Changes needed:
- [ ] Add support for both referral and postback support to handlers.py (https://github.com/nyaruka/rapidpro/blob/master/temba/channels/handlers.py#L2310)
- [ ] Adapt current Facebook referral triggers to make the referral id optional as a catchall (IE, if there isn't a trigger that catches that specific referrer id, then use the trigger with no referrer id, also validate there can only be one) (https://github.com/nyaruka/rapidpro/blob/master/temba/triggers/views.py#L297)
- [ ] When triggered, pass the actual referrer id to the triggered flow in `extra`. This lets a catch all referrer trigger split or save the referrer id for later use. (https://github.com/nyaruka/rapidpro/blob/master/temba/triggers/models.py#L255)

We'll also want to update the Facebook claim page to make sure people subscribe to `messaging_referrals` events so they get passed to us.

nicpottier:
@johncordeiro This would be a great task for you guys to take on!

cbunicef:
Happy to help with testing on this!

cbunicef:
Is it possible to build this as a universal handler for `extra`? There are other uses for information that's sent through postback (from the first message as well as later messages).

eg, FB passes info on ad attribution in the first postback:
https://developers.facebook.com/docs/messenger-platform/guides/ads#adattribution

norkans7:
@johncordeiro any progress on this?

johncordeiro:
@norkans7 next week we will create the PR for this.

cbunicef:
Hi all, any progress on this? We have a few good opportunities to use this coming up, the soonest next week.

johncordeiro:
Hey @cbunicef, it's in progress at the moment: https://github.com/ilhasoft/rapidpro/tree/feature/facebook-ref. The PR is coming ASAP.

cbunicef:
Is this ready to go? If so, how can I use it within RapidPro?

PR #605, opened 2017-07-24
Support for m.me referrals for Facebook

johncordeiro:
This PR adds support for m.me referrals for Facebook mentioned in issue #592.

Changes:

- [x] Add support for both referral and postback to handlers.py;
- [x] Facebook referral triggers are now optional (as a catchall);
- [x] When triggered, passes the actual referrer id to the triggered flow in extra;
- [x] Facebook claim page updated to make sure people subscribe to `messaging_referrals`.


norkans7:
Looks good after you fix the PEP8 errors

johncordeiro:
@norkans7 done!

norkans7:
100% Coverage too

temba/triggers/models.py
```diff
@@ -271,7 +275,11 @@ def catch_triggers(cls, entity, trigger_type, channel, referrer_id=None, extra=N
             triggers = triggers.filter(models.Q(channel=channel) | models.Q(channel=None))
 
         if referrer_id is not None:
-            triggers = triggers.filter(referrer_id=referrer_id)
+            triggers = triggers.filter(models.Q(referrer_id=referrer_id) | models.Q(referrer_id=''))
```
nicpottier: Realize this existed before but perhaps we should make the matching here be case insensitive, as that is probably confusing to users if we aren't matching just because of case.

johncordeiro: agreed, I also needed to update here: https://github.com/rapidpro/rapidpro/blob/master/temba/triggers/views.py#L317

temba/triggers/models.py
```diff
@@ -165,6 +165,10 @@ def archive_conflicts(self, user):
             else:
                 matches = matches.filter(groups=None)
 
+            # if this trigger has a referrer_id, only archive others with the same referrer_id
+            if self.referrer_id is not None:
+                matches = matches.filter(referrer_id=self.referrer_id)
```
norkans7: match the referrer_id case insensitevely here too

nicpottier:
@rowanseymour can we merge this guy?

PR #608, opened 2017-07-26
URN display

nicpottier:
* renames `urn` to `identity` on contact_urn, which contains the bits that determine equivalence
 * adds a `display` fields on urn
 * tweak lookups to be able to look up twitter urns both from screen names or ids
 * modify ChannelType to allow registering periodic tasks to run
 * modify TwitterChannelType to periodically remap Twitter keys
 * modify TwitterChannelType to send using the appropriate screen_name or id
 * modify handlers to create appropriate URN types


nicpottier:
Consideration:

How do we handle imports of say Twitter screen names? Do we immediately try to convert these to ids by trying to look them up? Looking at the APIs, it looks like we can grab these 100 at a time:
  https://dev.twitter.com/rest/reference/get/users/lookup

And the API limits are 900/15 mins so then 90k per 15 minutes, which isn't super fast for large imports. (or the upcoming migration)

So I guess the question is do we think it is valid to have a Twitter URN that does not yet have an id.. something like `twitter:#rapidpro`? That would allow us to backfill those lazily, but might lead to weird cases where say people add someone to a flow that uses quick replies and therefore needs the actual user id.

Maybe imports need to be "slowed down" such that they are not complete until all these ids are resolved? IE, once we do the migration, we never allow a Twitter URN to enter the system without an id set?

johncordeiro:
I think the best option is to slow down the import until we have all these ids in contacts and after that never allow twitter contacts without ids.

PR #612, opened 2017-08-01
Add Token Authentication to the Junebug handler

KaitCrawford:
This allows messages and events from Junebug to be authenticated if the channel has a token specified.


KaitCrawford:
@nicpottier please review. This adds optional token authentication onto messages and events coming from Junebug.

KaitCrawford:
@rowanseymour are you happy to merge this?

PR #613, opened 2017-08-02
PR for USSD fixes

erikh360:
Some of the fixes made on our side for USSD.

### DIFF
--- a/temba/channels/models.py
+++ b/temba/channels/models.py
@@ -211,6 +211,7 @@ class Channel(TembaModel):
     CONFIG_MAX_LENGTH = 'max_length'
     CONFIG_MACROKIOSK_SENDER_ID = 'macrokiosk_sender_id'
     CONFIG_MACROKIOSK_SERVICE_ID = 'macrokiosk_service_id'
+    CONFIG_RP_HOSTNAME_OVERRIDE = 'rp_hostname_override'
 
     ENCODING_DEFAULT = 'D'  # we just pass the text down to the endpoint
     ENCODING_SMART = 'S'  # we try simple substitutions to GSM7 then go to unicode if it still isn't GSM7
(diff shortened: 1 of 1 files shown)


temba/channels/models.py
```diff
@@ -211,6 +211,7 @@ class Channel(TembaModel):
     CONFIG_MAX_LENGTH = 'max_length'
     CONFIG_MACROKIOSK_SENDER_ID = 'macrokiosk_sender_id'
     CONFIG_MACROKIOSK_SERVICE_ID = 'macrokiosk_service_id'
+    CONFIG_RP_HOSTNAME_OVERRIDE = 'rp_hostname_override'
```
nicpottier: Can you talk me through this? Are you guys only exposing this endpoint or something?

temba/channels/models.py
```diff
@@ -1428,9 +1432,16 @@ def send_junebug_message(cls, channel, msg, text):
 
         if is_ussd:
             session = USSDSession.objects.get_session_with_status_only(msg.session_id)
-            external_id = Msg.objects.values_list('external_id', flat=True).filter(pk=msg.response_to_id).first()
-            # NOTE: Only one of `to` or `reply_to` may be specified
-            payload['reply_to'] = external_id
+            # make sure USSD responses are only valid for a short window
+            response_window = timedelta(seconds=180)
```
nicpottier: Might be slightly clearer to just calculate the oldest here instead of in the `if`. .

IE: 
```
response_expiration = timezone.now() - response_window
```

temba/channels/models.py
erikh360: @nicpottier In the Uganda data centre the routing to the hostname we have for RapidPro would go via the internet and because their bandwidth is so terrible we had lots of timeouts, so we needed to override the hostname to use an internal address

erikh360:
This was merged as part of https://github.com/rapidpro/rapidpro/pull/591

PR #620, opened 2017-08-22
Feature: resume or restart interrupted USSD pull sessions

### LINKED ISSUE #609: USSD: Support for resuming interrupted sessions (USSD Pull)
This issue is a follow-on to the discussion in #588 .

Currently, the system does not contain functionality to allow a user to resume an interrupted session on a USSD pull channel. The expected user flow would be:
1) Contact is active in a USSD pull flow
2) Contact's session times out or otherwise ends
3) Contact redials USSD number 
4) System detects that Contact was active in a flow (i.e. has not completed most recent flow)
5) System displays a menu offering the option to Continue the flow, or restart the flow (see attached screenshot for example)

phone

nullcode:
Closes #609

@nicpottier @ericnewcomer @rowanseymour Please review.


ericnewcomer:
@nullcode I'm not quite following this. Can you give more information on how resumption is modeled in the editor?

RichardHeslopZA:
@nullcode Could you please share some screenshots of how the resume functionality will be configured on the front-end? Thanks!

nullcode:
@ericnewcomer @RichardHeslopZA there's no editor changes here, it's implemented how @RichardHeslopZA specified it in #609. Every Pull session will automatically provide the resume/restart menu when the contact recalls the number.

ericnewcomer:
Still not quite following, if they are interrupted, won't they follow that route in the flow? How and where are they resumed when coming back in? Seems once they are interrupted they have no flow run to resume.

nullcode:
@ericnewcomer That's right, they will end up in a ruleset interrupt clause and that's what the system detects and picks that ruleset up and resume the run from there (basically reentering the ruleset again) or from the entry point if that's what the user decides with.
Here's a brief rundown:
1. user starts a flow
2. interrupt happens
3. ussd session and flow run are interrupted
4. user retriggers the flow by calling the same code
5. ussd session is picked up because it has been interrupted last time
6. flow run and steps are identified by the ussd session.
7. a menu is sent to user to resume or restart the flow

I've created two tests that are quite elaborate of describing the flow of actions (`test_resume_interrupted_pull_session` and `test_restart_interrupted_pull_session`)

ericnewcomer:
The problem here is that interruption and resumption are really two different things and should be modeled as such. The resume option really needs to be a distinct rule type that is placed underneath an interrupt rule. While this is verbose, it's the only way to ensure that the correct experience can be modeled for the end user.

Interruption is intended to clean up things (remove from groups, take any actions when a run in terminated and may never be resumed). A resumption should occur after an interruption is dealt with. If no interruption-specific actions need to be taken, then routing it directly to a "Wait for Resumption" rule is what you would need to do if you want to manage the resume case.

nullcode:
@ericnewcomer I'd thought about this too, but I don't necessarily agree that this ensures the correct behaviour. If you deal with interruption specific things, like removing someone in a group, does that mean you can't resume it afterwards? If you would still connect a "Wait for resumption" after those that would mean the same thing that it's doing now. If anything it should revert those actions. Another thing is that this "Wait for resumption" ruleset would be kind of a nightmare to implement since it has so many special things to do, like automatically picking up connections based on where you placed it. What if resumption would only work when you don't connect anything to the interrupt branch?

ericnewcomer:
There's no way to know the appropriate way to resume, which is why connecting to a wait for resume is the best we can do. For example, we don't know if showing the exact same menu is what the flow designer wants to happen in that case (perhaps they want to go to a different "welcome back" version of that menu or ride to a different part of the flow entirely).

Since we view routing the resume case as essential, we have to see interruption and resumption distinct things. A resumption is not required after a resume, but can't occur without an interruption first. It's up to the flow designer if they want to take action after interruption or not or wherever they want to undo those actions upon resumption. Yes, this can be cumbersome for the designer, but allows for flexibility to build the best possible end user experience. 

FWIW, we see this behavior as something that should likely be added to IVR as well to handle disconnects and reconnects. In the disconnect case there you might trigger a new call to resume the flow or start a different flow entirely.

nullcode:
I feel you and I understand that the only way to handle all possible cases is to pass all the responsibility to the flow designer (and in this case it really gets complicated for the designer to know how to do it properly too - imagine removing someone from groups then putting them back when the resumption happened and also duplicating these as many times as many choices they want to make available.)

I have the impression that the current feature solves the issue that @RichardHeslopZA came up with. I'd suggest them to fork this branch and try it out before we spend 2-3x more time implementing an editor based version.

ericnewcomer:
The issue is that we see this behavior as something that is also missing from IVR, so we have pretty strong opinions on how it should behave since it needs to have a common vision. We currently view this flexibility for the flow designer as pretty essential in order to model the proper behavior for interruption and resumption independently. That said, we are certainly open to other ideas that can meet that goal.

Forking and deploying this is certainly an option for those that need it near term, but just be aware that we're unlikely to merge it back in without addressing the concerns above.

nullcode:
@ericnewcomer The only idea I could come up with off the top of my head which might be a bit easier on the editor is to treat the actions connected to the interrupt branch differently, eg. having a checkbox on them saying "Revert upon resume", so the resumption could be automated without making a ton of connections and reverts from all the "wait for" rulesets. That being said it only works if we assume that the resumption is what should happen after every every re-trigger/interrupt. 
From my perspective, it's all about trying to make a comfortable automatic handling solution, and limiting users to a certain behaviour model versus a do-everything-on-your-own and understand all the implications which already feels like an Apple vs Android debate lol. At the end of the day I still treat RapidPro as a visual programming language, thus the elements should be as explicit as they can, but man sometimes I wouldn't wish it on anyone to create complex flow structures.

ericnewcomer:
Flows need to flow forward. We definitely cannot do anything that attempts to "unwind" actions or steps taken.

If we allow resume, it seems to me we just can't assume how the designer would want to handle that resumption. For example, they might need to flag the contact as having been interrupted, give them a different prompt explaining they been resumed, or put them in a different pool to control against for statistical surveys, or any number of other things we aren't thinking of..

Take the "Other" case as a similar example. We could have just repeated the question for simplicity or modeled the questions and answers together, but RP has always been opinionated on erring toward more flexibility for the flow designer in these scenarios to give the best possible experience for the end-user.

What bit of this in the editor do you see as a problem? Having a resume ruleset as an option doesn't cloud the common case I don't think.

nullcode:
Yeah that's what I meant with the limiting behaviour vs all around flexibility, but yeah the problem I see with the resumption ruleset is that it might end up in a mess of connections. Let's say you have X amount of menus and Y1...Yn actionsets connected to each interrupt branch. Let the resumption ruleset have Z menu options (likely to be 2, but maybe more). That means you need X amount of resume rulesets and Z (or more) times Y1...Yn actionsets with inverted actions (if you want to put users back to groups, remove label, etc.). Also would this occur to the flow designers at any point ? Would need some examples/tutorials for this for sure. From my point of view its a bit inconvenient is all, where as with sacrificing some flexibility it could be easier. That being said I understand your point that you don't want to unwind actions like that.

nullcode:
We literally implemented this based on user feedback and someone might come up with something like 'a resume needs to call a webhook on the moon,' but probably around 90% of the cases just want to resume where it was left off or start from the entry point, but enough rambling from me on this.

nicpottier:
Totally understand the complexity side of things, but really the big thing that just cannot change is going back in a flow, that's just a no-go from an architecture point of view.

As Eric said, we are open to other implementations if you have other ideas, but we can't just rewind a flow to a previous action, that's not a paradigm we want to support and we think that would serious long term consequences.

Yes there is more complexity in explicitness, but there is also the ability to create much better user experiences for contacts by doing so, and in the end a few more minutes of users dragging stuff around is far better to guarantee that better experience for thousands of contacts.

RichardHeslopZA:
Perhaps also to add: the resume menu would need to support translations. i.e. would somewhere/somehow need to be surfaced in the designer with translation capabilities.

nicpottier:
Closing, needs redesign which will be discussed here:
https://github.com/rapidpro/rapidpro/issues/712

PR #629, opened 2017-09-07
Removes giant join through run/flow to figure out flow steps

nicpottier:
That was leading to some very very poor query performance.


norkans7:
why not this PR on nyaruka fork?

PR #631, opened 2017-09-08
Fix Chatbase integration

teehamaral:
This PR fixes the running on the flow simulator when there a Chatbase account connected to the organization.


temba/msgs/models.py
```diff
@@ -881,7 +881,7 @@ def process_message(cls, msg):
 
         # Sending data to Chatbase API
         (chatbase_api_key, chatbase_version) = msg.org.get_chatbase_credentials()
-        if chatbase_api_key:
+        if chatbase_api_key and msg.channel:
```
rowanseymour: I think better to be more explicit and check `msg.contact.is_test` before even calling `msg.org.get_chatbase_credentials()`

temba/msgs/models.py
```diff
@@ -880,8 +880,10 @@ def process_message(cls, msg):
             chatbase_not_handled = False
 
         # Sending data to Chatbase API
-        (chatbase_api_key, chatbase_version) = msg.org.get_chatbase_credentials()
-        if chatbase_api_key and msg.channel:
+        (chatbase_api_key, chatbase_version) = msg.org.get_chatbase_credentials() if not msg.contact.is_test \
+            else (None, None)
+
+        if chatbase_api_key:
```
rowanseymour: Seems simpler to write this as ...

```python
if not msg.contact.is_test:
  (chatbase_api_key, chatbase_version) = msg.org.get_chatbase_credentials()
  if chatbase_api_key:
    ...
```

no?

teehamaral: Yes, I'll change

PR #632, opened 2017-09-09
Handle 4XX errors for flow management

### LINKED ISSUE #616: Put error body into @extra for further processing on endpoints that return helpful failure detail
Looking to find a solution for hitting endpoints that don't always return a 200, yet that 4XX is actionable within flows.  For example, hitting the endpoint with missing keys in the query string parameters technically should not return a 200 with a dictionary indicating a failed status/error message.  We've essentially done this, but it breaks the web and results in mistakes as developers try to _Do The Right Thing_ when building endpoints for RapidPro.

I think that the logging of the error needs to begin [here](https://github.com/rapidpro/rapidpro/blob/3d9008da2eaddbf6010dd73e177781e3ccd3682c/temba/api/models.py#L328-L331) but I'm not sure how one would go about sticking the response into `@extra`.

Appreciate the help!

/cc @nicpottier @putorti

chris-erickson:
We have a problem with returning _failures_ that can be acted on.  4XX class errors could potentially be managed and resolved.  Right now we have to return a 200 that provides a failure code as part of the body so we're all to check it.

I'm not entirely this is the right way to do this, but it's a start.

fixes #616

### DIFF
--- a/temba/api/models.py
+++ b/temba/api/models.py
@@ -311,19 +311,20 @@ class WebHookEvent(SmartModel):
                 body = 'Skipped actual send'
                 status_code = 200
 
-            if (200 <= status_code < 300) or (400 <= status_code < 500):
-                try:
-                    response_json = json.loads(body, object_pairs_hook=OrderedDict)
+            # process the webhook response
+            try:
+                response_json = json.loads(body, object_pairs_hook=OrderedDict)
 
-                    # only update if we got a valid JSON dictionary or list
-                    if not isinstance(response_json, dict) and not isinstance(response_json, list):
-                        raise ValueError("Response must be a JSON dictionary or list, ignoring response.")
+                # only update if we got a valid JSON dictionary or list
+                if not isinstance(response_json, dict) and not isinstance(response_json, list):
+                    raise ValueError("Response must be a JSON dictionary or list, ignoring response.")
 
-                    run.update_fields(response_json)
-                    message = "Webhook called successfully."
-                except ValueError:
-                    message = "Response must be a JSON dictionary, ignoring response."
+                run.update_fields(response_json)
+                message = "Webhook called successfully."
+            except ValueError:
+                message = "Response must be a JSON dictionary, ignoring response."
 
+            if 200 <= status_code < 300:
                 webhook_event.status = cls.STATUS_COMPLETE
             else:
                 webhook_event.status = cls.STATUS_FAILED
--- a/temba/api/models.py
+++ b/temba/api/models.py
@@ -311,7 +311,7 @@ class WebHookEvent(SmartModel):
                 body = 'Skipped actual send'
                 status_code = 200
 
-            if 200 <= status_code < 300:
+            if (200 <= status_code < 300) or (400 <= status_code < 500):
                 try:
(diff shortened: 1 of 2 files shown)


temba/api/models.py
```diff
@@ -311,7 +311,7 @@ def trigger_flow_event(cls, run, webhook_url, node_uuid, msg, action='POST', res
                 body = 'Skipped actual send'
                 status_code = 200
 
-            if 200 <= status_code < 300:
+            if (200 <= status_code < 300) or (400 <= status_code < 500):
```
nicpottier: I don't think this is right, you are now making 4** drop into the success case instead of the else as they were before.

Can either remove the else entirely and keep the attempt for json parsing in this block the same then set webhook_event.status to `COMPLETE` or `FAILED` based on the status (and throw appropriately) or repeat the attempt to parse the json and update the run in the else case. Former is probably a bit cleaner.

chris-erickson:
@nicpottier thanks for the direction.  I took another run at it.  I'm a bit unclear how everything works together - since we want to not treat it as "failed", in the sense that we want to still act on the response.

I presume that `webhook_event.status = cls.STATUS_FAILED` is just a status, but doesn't prevent examining what the response contains.  Before this PR, a 500 error would just pass the entire parsing step and `run.update_fields(response_json)` would never happen, hence the response isn't available for examination in RP.  Do I have that right?

nicpottier:
Nice, I think this looks good. One last thing, can you add a test case to `test_webhook` in `flows/tests.py`? (around line 4185 or so) (!!) You can just change the payload of the mockresponse returning 500 to be JSON and then check that it was set in extra? (see earlier test cases for examples)

chris-erickson:
@nicpottier I added a test, but as you alluded to, the test are quite.. extensive.  I wasn't able to locate the failure but I suspect I have an assert a bit off for this case.  Hopefully you can spot it quick.

temba/flows/tests.py
```diff
@@ -4180,6 +4180,20 @@ def test_webhook(self):
...
             self.assertEquals("1001", incoming.text)
 
+        with patch('requests.post') as mock:
+            mock.return_value = MockResponse(400, '{ "error": "400", "message": "Missing field in request" }')
+            rule_step.run.fields = None
+            rule_step.run.save()
+
+            (match, value) = webhook.find_matching_rule(webhook_step, run, incoming)
+            (match, value) = rules.find_matching_rule(rule_step, run, incoming)
+            self.assertIsNone(valid_uuid, match.uuid)
```
norkans7: @chris-erickson 
First `assertIsNone` takes one argument, I think here you need `assertEquals` 

And you need to make sure you have a rule that will be matched on the ruleset https://github.com/rapidpro/rapidpro/pull/632/files#diff-0f24266e2d8c853528ec9af33da0614bR4030

so valid_uuuid is the rule when the response contains Valid

the other rule matches when we have 'Invalid'

1. You need a predictable uuid for the invalid rule 
2. You need invalid in the response or better a new rule in the ruleset that

chris-erickson:
Yay, prodding the black box was successful!

Will you handle merge from here?

nicpottier:
Yep, we'll merge and ship this to resist sometime in the next few days.

PR #646, opened 2017-10-05
New Skype Bot Channel

gilsonbp:
Implementing the new channel by integrating with Skype using Bot Framework
https://docs.microsoft.com/en-us/bot-framework/portal-register-bot


rowanseymour:
@gilsonbp this will be neat to have merged - but we'll need it updated to use the dynamic channel type system in master.  Please take a look the classes in temba/channels/types/ for examples of what these look like. We're basically moving all type specific code in their own modules. The exception is the handler code in temba/channels/handlers.py. We haven't moved that because eventually all handler code will be moved to another project called [courier](https://github.com/nyaruka/courier), so for now it's fine to leave the SkypeHandler in there.

rowanseymour:
Closing this due to lack of activity

PR #649, opened 2017-10-12
Run lint on Python 3

alexmuller:
This pull request involves splitting the Travis build into multiple parts so that it's possible to run just the lint on Python 3.

It may be possible to split the build step further into more steps but I'm not familiar enough with the project yet to know.

This will keep the codebase Python 3-friendly and allow fixing the tests gradually over time.

## Example Travis output

screenshot from 2017-10-12 17-36-47

### DIFF
--- a/temba/channels/courier.py
+++ b/temba/channels/courier.py
@@ -39,7 +39,7 @@ def msg_as_task(msg):
     Used to serialize msgs as tasks to courier
     """
     msg_json = dict(id=msg.id,
-                    uuid=unicode(msg.uuid) if msg.uuid else "",
+                    uuid=str(msg.uuid) if msg.uuid else "",
                     org_id=msg.org_id,
                     channel_id=msg.channel_id,
                     channel_uuid=msg.channel.uuid,
--- a/temba/channels/handlers.py
+++ b/temba/channels/handlers.py
@@ -2364,7 +2364,7 @@ class JioChatHandler(BaseChannelHandler):
         create_time = body.get('CreateTime', None)
         msg_date = None
         if create_time:
-            msg_date = datetime.utcfromtimestamp(float(unicode(create_time)[:10])).replace(tzinfo=pytz.utc)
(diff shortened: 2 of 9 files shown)


temba/channels/courier.py
```diff
@@ -39,7 +40,7 @@ def msg_as_task(msg):
     Used to serialize msgs as tasks to courier
     """
     msg_json = dict(id=msg.id,
-                    uuid=unicode(msg.uuid) if msg.uuid else "",
+                    uuid=text_str(msg.uuid) if msg.uuid else "",
```
rowanseymour: elsewhere we've started using six.text_type for this, tho it seems like it's always safe to use `str` for something like a UUID, no?

alexmuller: Ah yeah I guess that makes sense for a UUID. I'm still very new to RapidPro so I might be being a bit overly cautious. Thanks!

temba/channels/handlers.py
```diff
@@ -2364,7 +2365,7 @@ def post(self, request, *args, **kwargs):
         create_time = body.get('CreateTime', None)
         msg_date = None
         if create_time:
-            msg_date = datetime.utcfromtimestamp(float(unicode(create_time)[:10])).replace(tzinfo=pytz.utc)
+            msg_date = datetime.utcfromtimestamp(float(text_str(create_time)[:10])).replace(tzinfo=pytz.utc)
```
rowanseymour: `six.text_type`

temba/channels/types/junebug/type.py
```diff
@@ -93,7 +94,7 @@ def send(self, channel, msg, text):
             event.response_body = response.text
 
         except Exception as e:
-            raise SendException(unicode(e), event=event, start=start)
+            raise SendException(text_str(e), event=event, start=start)
```
rowanseymour: `six.text_type`

alexmuller:
Updated, thanks!

rowanseymour:
Looks good I think. I guess that Python 3.6 bit will be failing for a while. 

@nicpottier you foresee any issues with this?

Issue #607, opened 2017-07-26
Add action for blocking and stopping contacts


nicpottier:
So people can build flows and hook up keywords to stop people themselves.

(same thing for block?)

Right now you can sort of do something similar by removing people from all groups which is essentially the same thing, but would be nice to keep this consistent.

teehamaral:
Is there something like in progress now? If not, would be welcome a PR?

nicpottier:
Ya, we've talked about this various times. I think the `stop` case is less clear that it would make sense, as stopping is something the contact does, not the administrator. Blocking is more the type of action that administrators do on a contact if they decide not to send to them, so I think that action could make sense. (unlike stop)

I think the only tricky bit there is thinking through what it means to block a contact that is currently in a flow. It seems to be we would probably want to block someone from adding steps to a flow after they have blocked a contact, as we won't send messages (nor should that contact be able to be in a flow after being blocked)

So this would be more complicated than just adding the action, we'd also need to add the editor tweaks to make sure we didn't allow extending the flow beyond this action.

All that to say yes, we would accept a PR but it isn't as simple as it looks. :)

teehamaral:
Yep, I'll open a PR with an initial code soon, so we can discuss it there, right?

nicpottier:
Yep yep.

nicpottier:
Related: #1241

rowanseymour:
Added as set_contact_status action in engine - part of Update Contact in editor

PR #650, opened 2017-10-13
Action for stopping contact

teehamaral:
This PR includes an action to stop contact on flow #607


teehamaral:
@nicpottier I have thought about something like, what do you think?

nicpottier:
I'm confused, did you not ready my comment? Stop is the wrong action here, blocking is more correct and changes need to be made to not allow continuation of the flow after it.

PR #653, opened 2017-10-17
Upmerge from Nyaruka

rowanseymour:
More dynamic channels etc


nicpottier:
let's merge the changelog from 213-216 to just 216?

PR #659, opened 2017-10-19
Feature/nlu api (Bothub and Wit.AI)

vctrferreira:


nicpottier:
He guys.. seriously.. 

1) You need a description for this PR, what is it exactly?
2) You need to open issues and talk about adding features and PRs before submitting PRs

Not all PRs will be accepted, especially if you don't chat with us about them first. We may choose to just let this live in your brach instead. But at least have a conversation before opening these. 

Please add a description to this, why you are adding it, what it provides, what your approach is and detail the thought you've put into as to why this can't be solved via normal webhooks or other approaches.

vctrferreira:
Sorry, I was wrong about the repository, it is to ilhasoft/rapidpro.

PR #660, opened 2017-10-22
Enforce modern assertEqual() syntax

alexmuller:
`assertEquals()` is deprecated in Python 2 and disappears in Python 3.

Sorry about the size of this diff, I understand if this isn't a priority for you.

I'm trying to make changes in such a way that it makes the codebase more likely to run on Python 3 without being too disruptive. In this case that means adding a new dependency which will prevent any further additions of the deprecated `assertEquals()` syntax.

### DIFF
--- a/pip-freeze.txt
+++ b/pip-freeze.txt
@@ -17,7 +17,6 @@ certifi==2017.4.17        # via python-telegram-bot, requests
 cffi==1.10.0              # via cryptography
 chardet==3.0.4            # via requests
 colorama==0.3.9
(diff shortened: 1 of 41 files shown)


nicpottier:
Bug +1 from us on this, very happy to have some help on bringing things to Py3. (it isn't that we don't want this, just other fish to fry right now so the help is highly appreciated!)

alexmuller:
Thanks very much for reviewing this, glad it's helpful!

PR #662, opened 2017-10-27
Pathfix for highcharts

cromulus:
the slash at the front breaks the link.


jlev:
👍 

the extra slash breaks highcharts if you're loading static files from S3

PR #739, opened 2018-03-23
Include organization and revision IDs in flow webhook events

jeremyist:
## Changes

This PR includes a quick adjustment to the flow webhook to automatically include organization info and the flow's revision ID in the JSON payload. 

## Use Cases

- Including organization info in the payload would make it trivial to use common webhook endpoints in multi-tenant rapidpro installs; without the need for users to input org-specific urls/headers (less risk of human error/mixing things up). 

- Including the revision ID allows for more precise analysis of flow results over time, as the flows are fine-tuned. A/B testing of sorts. 

## Notes

Adding org info was pretty straightforward. Revision id proved to be slightly trickier - I ended up settling on using the metadata dict as it seemed to safely contain the needed info. 

Always open to suggestions!

### DIFF
--- a/media/test_flows/color.json
+++ b/media/test_flows/color.json
@@ -7,6 +7,7 @@
       "flow_type": "F",
       "base_language": "base",
       "metadata": {
+        "revision": 12,
         "uuid": "2a385c5b-e27c-43ac-bbc6-49653fede421",
         "author": "Ryan Lewis",
         "name": "Color Flow"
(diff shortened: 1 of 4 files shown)


jeremyist:
Thanks for the quick feedback @rowanseymour - I made the adjustment to revision as requested. 

As far as orgs go - I can definitely think of some use cases for including it, but I can also understand why it may not be preferable to include it (particularly if it would be going away again in the near future). Let me know for sure, either way. Thanks again!

rowanseymour:
@jeremyist org is something we might be open to adding in future, and it's worth mentioning that in the new engine - users can define the entire webhook body with an expression, but for now i'd rather just merge the flow revision change without adding org.

PR #742, opened 2018-03-30
Include flow/revision info in sms events

jeremyist:
This PR includes an adjustment to  `trigger_sms_event` - if a message is tied to a flow  this will expose that information in the webhook payload. For similar reasons as with #739  , revision is included as well. 

Currently the webhook will only show if the event was handled via `status` - but it won't show the context in which it was handled. Adding that information makes the webhook tremendously more powerful for analytics purposes.

### DIFF
--- a/temba/api/models.py
+++ b/temba/api/models.py
@@ -367,6 +367,13 @@ class WebHookEvent(SmartModel):
                     status=msg.status,
                     direction=msg.direction)
 
+        step = msg.steps.all().first()
+
+        if step:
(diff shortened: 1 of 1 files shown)


nicpottier:
The ability to plug in global webhooks is being deprecated, it is a poor model of integration as it makes us wait for the outside world, which invariably leads to melting people as we go up in scale which then is something for us to manage. People should be polling the API endpoint to get changes instead or using in-flow webhooks to handle real-time callbacks.

PR #757, opened 2018-05-28
Modified the key used in the transferto API call

chopley:
Currently str(int(time.time()) returns the same key for events requested on the same second. 

As per TransferTo documentation this will result in API rejection since the transaction is rejected if the key has already been used. 

This could easily be fixed with str(int(time.time()*1000)) as suggested in the documentation Ch 2.3 https://shop.transferto.com/shop/v3/doc/TransferTo_API.pdf


temba/airtime/models.py
```diff
@@ -53,7 +53,7 @@ def post_transferto_api_response(cls, login, token, airtime_obj=None, **kwargs):
         if not settings.SEND_AIRTIME:
             raise Exception("!! Skipping Airtime Transfer, SEND_AIRTIME set to False")
 
-        key = str(int(time.time()))
```
rowanseymour: Can we use something genuinely random here? even a uuid?

rowanseymour:
@chopley if you merge in latest master it should fix the failing tests here

temba/airtime/models.py
chopley: Judging from the requirement of a distributed server (see below) and reading between the lines, I think it needs to be an increasing integer?

> The following is an example of a	good key.
> 1. Use a unique server: 
> a time stamp with milliseconds.
> Use multiple servers.	
> Use a common sequence shared through a database to ensure that all the servers are synchronized.	
> Set up several accounts through the Transfer-To web interface to have distinct sequences

norkans7: right the key needs to be an integer 👍

chopley: Judging from the suggestions for multiple servers, it might also require that successive calls to the TransferTo API use _increasing_ integers for the key. I'm not sure if that is something worth worrying about for RapidPro use on distributed servers? Probably not, but might be worth suggesting that users not use the same TransferTo accounts in that instance.

chopley: Cool thanks!- 
The reason I think that the integers need to be increasing is because the mention sharing a sequence if using multiple servers. I can only imagine this being necessary if you need to know what the previous number was, and hence why I _think_ they _might_ need to be increasing.

A timestamp for multiple servers won't be good in this case  since the individual clocks might not be 100% synchronous, and you might derive it from a machine where the clock is slightly delayed compared to another.

Above are just my intuition! :) 

Thanks for all the feedback.

nicpottier: If we are worried about clock drift across different servers, then it seems using a Redis key with an INCR might make the most sense here.

chopley: That would work (and is in-line with their recommendation i.e. "Use a common sequence shared through a database to ensure that all the servers are synchronized."). I'm probably not the most well placed to actually implement this though as my coding skills are that of a data scientist :) 

I could have a go, but I suspect it is a probably a couple of lines for someone more experienced with Redis and how it operates within RapidPro?

norkans7: You can open an issue on https://github.com/rapidpro/rapidpro/issues may be with a link to this discussion

nicpottier: Realize I'm back pedaling but we may be bike shedding a bit more than we need to. I'm guessing in reality millis is good enough for any use case we are likely to see. (and doesn't then introduce the dependency on Redis which might get harrier from a GoFlow perspective)

I'm fine merging this with millis.

Issue #760, opened 2018-06-15
Allow basic authentication for API


nicpottier:
Some external tools don't allow using Token authentication as we use in our API. Might be nice also allow basic authentication using API tokens. IE, username is the email address of the user, password is the API token, org is derived from the pair. Should be an easy plugin.

rowanseymour:
Added in #761

PR #761, opened 2018-06-18
adding HTTP Basic Authentication for API

xkmato:
This is a response to the call for help on issue #760 

The logic is entirely the same as Token Authentication logic with one extra step of verifying if the token owner has the same username as the one provided in the credentials.
 Let me know what you think

### DIFF
--- a/temba/api/support.py
+++ b/temba/api/support.py
@@ -55,10 +55,10 @@ class APIBasicAuthentication(BasicAuthentication):
         try:
             token = APIToken.objects.get(is_active=True, key=password)
         except APIToken.DoesNotExist:
-            raise exceptions.AuthenticationFailed("Invalid token")
+            raise exceptions.AuthenticationFailed("Invalid token or email")
 
         if token.user.username != userid:
-            raise exceptions.AuthenticationFailed("Invalid username")
+            raise exceptions.AuthenticationFailed("Invalid token or email")
 
         if token.user.is_active:
             # set the org on this user
@@ -66,7 +66,7 @@ class APIBasicAuthentication(BasicAuthentication):
 
             return token.user, token
 
-        raise exceptions.AuthenticationFailed("Invalid token")
+        raise exceptions.AuthenticationFailed("Invalid token or email")
 
 
 class OrgRateThrottle(ScopedRateThrottle):
--- a/temba/api/support.py
+++ b/temba/api/support.py
@@ -1,7 +1,8 @@
 import logging
 
+from django.contrib.auth.models import User
 from rest_framework import exceptions, status
-from rest_framework.authentication import TokenAuthentication
+from rest_framework.authentication import TokenAuthentication, BasicAuthentication
 from rest_framework.exceptions import APIException
 from rest_framework.renderers import BrowsableAPIRenderer
 from rest_framework.throttling import ScopedRateThrottle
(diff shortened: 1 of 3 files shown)


temba/api/support.py
```diff
@@ -41,6 +42,33 @@ def authenticate_credentials(self, key):
...
+        if token.user.username != userid:
+            raise exceptions.AuthenticationFailed("Invalid username")
+
+        if token.user.is_active:
+            # set the org on this user
+            token.user.set_org(token.org)
+
+            return token.user, token
+
+        raise exceptions.AuthenticationFailed("User inactive or deleted")
```
nicpottier: These should all return "Invalid Token" otherwise you are leaking information. (like that the username doesn't exist in the system for example)

temba/api/v2/tests.py
```diff
@@ -266,34 +267,63 @@ def api_request(endpoint, token):
                 HTTP_AUTHORIZATION="Token %s" % token,
             )
 
+        def api_request_basic_auth(endpoint, username, token):
+            encoded = "Basic {}".format(base64.encodebytes("{}:{}".format(username, token).encode()).decode()
+                                        .replace('\n',''))
```
nicpottier: Maybe break out some variables here so that it's a bit easier to grok?

temba/api/support.py
nicpottier: (can change the TokenAuthentication above to do the same thing)

temba/api/v2/tests.py
xkmato: Changes made. Please check them out. Cheers

temba/api/support.py
```diff
@@ -38,7 +39,34 @@ def authenticate_credentials(self, key):
...
+    """
+
+    def authenticate_credentials(self, userid, password):
+        try:
+            token = APIToken.objects.get(is_active=True, key=password)
+        except APIToken.DoesNotExist:
+            raise exceptions.AuthenticationFailed("Invalid token")
+
+        if token.user.username != userid:
+            raise exceptions.AuthenticationFailed("Invalid username")
```
norkans7: I think this still leak some info, may be error message can be "Authentication failed" in all places

norkans7:
@xkmato FWIW the build failure is caused by black code style check error 
If you run `black --line-length=119` on the changed files they will be adjusted
or add the git hooks https://github.com/rapidpro/rapidpro/blob/master/pre-commit.sh https://github.com/rapidpro/rapidpro/blob/master/post-commit.sh

temba/api/support.py
dodobas: I agree, we should log the actual reason why authentication failed, and return generic message to the API consumer

temba/api/v2/tests.py
```diff
@@ -263,7 +264,16 @@ def api_request(endpoint, token):
                 endpoint + ".json",
                 content_type="application/json",
                 HTTP_X_FORWARDED_HTTPS="https",
-                HTTP_AUTHORIZATION="Token %s" % token,
+                HTTP_AUTHORIZATION="Token {}".format(token),
+            )
+
+        def api_request_basic_auth(endpoint, username, token):
+            credentials_base64 = base64.encodebytes("{}:{}".format(username, token).encode()).decode()
```
dodobas: Please use Python3.6 f-strings formater (https://docs.python.org/3/reference/lexical_analysis.html#f-strings)

temba/api/support.py
xkmato: Personally, I would prefer to give the API consumer an informative error message so that they can easily tell where they made a mistake. And also since the Token is random and very hard to memorize, I think it would make sense for the user know if the authentication mistake is a result of providing the wrong Token or a result of providing a wrong email address.
I am wondering if it makes sense to instead change both error messages to `Invalid Email or Token` instead of the more generic `Authentication failed`?

temba/api/v2/tests.py
```diff
@@ -263,7 +264,16 @@ def api_request(endpoint, token):
...
                 content_type="application/json",
                 HTTP_X_FORWARDED_HTTPS="https",
-                HTTP_AUTHORIZATION="Token %s" % token,
+                HTTP_AUTHORIZATION=f"Token {token}",
+            )
+
+        def api_request_basic_auth(endpoint, username, token):
+            credentials_base64 = base64.encodebytes(f"{username}:{token}".encode()).decode()
+            return self.client.get(
+                endpoint + ".json",
```
dodobas: This should also be an f-string expression

temba/api/support.py
norkans7: Probably a more descriptive message like `Authentication failed, please check the email and token` would be enough
The main thing is to keep the same error message for all the failures and not leak information whether the user was deleted or the token expired

nicpottier:
Thanks Kenneth, looks good!

PR #799, opened 2018-09-06
Novo Technology Channel

johncordeiro:
This PR adds the [Novo Technology](http://www.novotechnologyinc.com/) SMS channel. NOVO Technology is an SMS Aggregator from the Republic of Trinidad and Tobago.

The channel was created following the documentation provided in direct conversations with Technical Team from the company.


nicpottier:
Looks good to me, @norkans7 can you take a gander as well?

temba/channels/types/novo/tests.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
```
norkans7: We now support Python 3, we no longer need __future__ import and six library

temba/channels/types/novo/type.py
```diff
@@ -0,0 +1,48 @@
...
+    courier_url = r"^nv/(?P<uuid>[a-z0-9\-]+)/(?P<action>receive)$"
+
+    code = "NV"
+    category = ChannelType.Category.PHONE
+
+    name = "Novo"
+
+    claim_blurb = _(
+        """If you are based in Trinidad & Tobago, you can purchase a short code from <a href="http://www.novotechnologyinc.com/">Novo</a> and connect it in a few simple steps."""
+    )
```
nicpottier: Pretty sure this replaces the claim template @norkans7

norkans7: That is show on the section for the channel type when we have all types listed,

the claim template is for the form

norkans7: Ohh right, the template is need for those that extends the AuthenticatedForm.

temba/channels/types/novo/type.py
```diff
@@ -0,0 +1,46 @@
+
+import six
```
norkans7: I think six is not needed too

PR #803, opened 2018-09-11
Fix referral parameters for Facebook Ads requests

johncordeiro:
When a new referral event is created for Facebook Ads requests, RapidPro gets the first referral trigger available and starts it. This is because of `referrer_id` [validation](https://github.com/rapidpro/rapidpro/blob/master/temba/triggers/models.py#L337) in the catch_triggers.

I suggest keeping `courier` creating channel events for ads requests, but fixing this function to avoid starting flows. 

Check tests to understand better.


johncordeiro:
@nicpottier @norkans7 @rowanseymour Do you agree with this approach to solving this problem or do you have any other suggestion for that?

norkans7:
I think the flow is started because on RapidPro the trigger has an empty referral_id
by removing that trigger without referral_id then the flow will be started only for the referral id specified on RapidPro

nicpottier:
Seems we should honor specific referral id triggers first over generic ones though no?

johncordeiro:
It's starting flows for trigger *with* referral id as well, actually currently is starting arbitrary triggers with any referral id configured for ads requests. Check in the tests that the referral was `join`, but the ad requested started the flow anyway.

In my opinion, ads requests should not start any flow because of its unexpected behavior for users.

nicpottier:
Well its starting them because there is a catch all referral trigger right? ads are considered referrals.

johncordeiro:
No, there is no catch-all referral trigger, and it's starting anyway.

norkans7:
If I get it well, when no referral_id in extra we should only match the referral trigger if it does not specify the referral and now we match any first referral trigger.

I think https://github.com/rapidpro/rapidpro/blob/master/temba/triggers/models.py#L337 needs an else block for that

nicpottier:
@johncordeiro perhaps part of the confusion is the PR only has a test? Do you have the fix you are proposing as well?

nicpottier:
Just so I fully understand this, do ads also have referral ids? Are they blank here because we are losing some information at the courier level? (agree seems we are matching where we shouldn't be but maybe this is an opportunity to make ads more useful as well)

johncordeiro:
Facebook just sends `referrer_id` on ads requests if it was configured in the ad creation, but it's not required, see the [docs](https://developers.facebook.com/docs/messenger-platform/discovery/ads#adattribution).

And yes, the PR was created to show the existing problem and also think in how to better use these ads requests without `referrer_id`. Check it now.

temba/triggers/models.py
```diff
@@ -340,6 +340,8 @@ def catch_triggers(cls, entity, trigger_type, channel, referrer_id=None, extra=N
             # if we catch more than one trigger with a referrer_id, ignore the catchall
             if len(triggers) > 1:
                 triggers = triggers.exclude(referrer_id="")
+        else:
+            triggers = triggers.filter(referrer_id="")
```
norkans7: We need to consider other trigger types with referrer_id null

johncordeiro:
@nicpottier can you approve that?

PR #804, opened 2018-09-21
Generate token for Novo Channel

johncordeiro:
This PR adds a Secret Key to Novo Channel, which is required to receive incoming messages on Courier.


nicpottier:
@johncordeiro just checking in on this.

PR #853, opened 2019-01-16
Start connections

nicpottier:
Per conversation here: https://github.com/nyaruka/mailroom/issues/59

Makes `created_by` and `modified_by` nullable so that `trigger_session_events` can create FlowStarts  when encountered in Mailroom.

Also adds an M2M to FlowStart to `channels_channelconnection` so that on retries / throttling we can properly look up what flow to start, extra to use etc..

Snuck in a JSON tweak for the mailroom task start schema since that's needed for IVR.


nicpottier:
arr, wrong repo

PR #918, opened 2019-05-02
Update django.po

JesusEkie:
An improved translation to better reflect the original message.

### DIFF
--- a/locale/fr/LC_MESSAGES/django.po
+++ b/locale/fr/LC_MESSAGES/django.po
@@ -49,11 +49,11 @@ msgstr ""
 
 #: temba/api/models.py:135
 msgid "The URL that we will call when our ruleset is reached"
-msgstr ""
+msgstr "L'URL que nous appellerons lorsque notre ensemble de règles sera atteint"
 
 #: temba/api/models.py:588 temba/channels/models.py:2445
 msgid "Time it took to process this request"
-msgstr ""
+msgstr "Temps requis pour le traitement de cette demande"
 
 #: temba/api/v1/serializers.py:295
 #, fuzzy, python-format
@@ -64,7 +64,7 @@ msgstr "Groupe non valide ou Contact ID"
 #: temba/api/v1/serializers.py:306
 #, python-format
 msgid "Unable to find contact group with uuid: %s"
-msgstr ""
+msgstr "Impossible de trouver le groupe de contacts avec l'uuid: %s"
 
 #: temba/api/v1/serializers.py:334
 #, fuzzy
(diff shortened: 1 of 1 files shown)


norkans7:
Hey @JesusEkie can you make the translation on Transifex?

https://www.transifex.com/rapidpro/rapidpro/

JesusEkie:
Hey @norkans7 

After sleepless nights, translating RapidPro into French today, I received a message from Transifex stating that someone had downloaded an old version of the French translation file. All my translation work for RapidPro in French disappeared while I was about to fully translate all the fields.

You can check the activity of Transifex to confirm it. I really do not know why my translation was simply deleted.

nicpottier:
Hi @JesusEkie, don't think anything has been lost. Just the contrary I've been playing with translations to make sure your work doesn't go to waste. There's a few issues I've been working to resolve:

1. The PO files themselves for RapidPro were far out of date from the real code. These files are generated by `makemessages` and that hadn't been run for years, so even a fully translated PO file would only affect some pages and a great things were missing.

2. The PO files on Transifex were also out of sync with the RapidPro ones, so we were doubly out of date.

3. The MO files which are actually used by Django to display translations were out of date because `compilemessages` wasn't run regularly.

I've addressed 2) by making sure Transifex is now integrated with Github so will always pull the latest PO files as they are updated.

I'm working on making 1) and 3) part of the build process so that builds fail if we have out of date PO or MO files to makes sure we don't fall behind as we have. I'm working on that here: 
   https://github.com/rapidpro/rapidpro/pull/925

I should be done with that today or tomorrow at the latest, at which time we should have updated PO and MO files and all translations you do should be valid and appropriate. I'm pretty sure we will still be able to merge in your PO file with Transifex and most of those translations will take at that time, but there will certainly be many that aren't translated because the PO file you were using was out of date.

I apologize for that and want to thank you for both taking on doing some translations and for motivating us to do better with these locale files and putting in a process that will guarantee these stay up to date in the future.

I'll update here when I have more.

JesusEkie:
Thank you @nicpottier for you answer.
You guy owe me a cup of coffee :-)

While you are working on it can I proceed to U-Report?
I saw that the French version contains untranslated fields.

nicpottier:
Ok, finally got that build working and updated `django.po` and `djangojs.po` (those will now be guaranteed to be updated when code changes, as will the `mo` files)

Merged to master and it looks like Transifex picked those up and merged them correctly. So I think if you do the translation there then once they are 100% reviewed Transifex will open a PR for them to be merged. 

Thanks!

Note that U-Report has the same problems as RapidPro (po files out of date, I'll try to do the same with the travis build and Transifex integration tomorrow)

PR #919, opened 2019-05-02
Update djangojs.po

JesusEkie:
An improved translation to better reflect the original message.


nicpottier:
As we chatted this should come from the Transifex side (which is now updated)

Issue #770, opened 2018-06-27
Call Webhook: Dynamic Value in Header Value


qoda:
Is it possible to use a dynamic value in the Header value. I.e. I call a authentication endpoint in my API to retrieve an auth token. The token is than used for subsequent webhook calls.

<img width="658" alt="screen shot 2018-06-27 at 09 28 17" src="https://user-images.githubusercontent.com/125271/41959168-81b88eaa-79ec-11e8-9d36-aa34c613197b.png">

rowanseymour:
Not currently possible but can't think of a reason we shouldn't add support for this.

qoda:
@rowanseymour will look into it and see if I can submit a PR for your consideration, it would be exceptionally useful to me.

qoda:
@rowanseymour I looked at the code last night to see if this would be possible, but have little to no clue where to even start. Will use a default token and see if the same can be achieved. I'll try and look at it again, but for now... deadlines. You can tag this as a feature request if you like.

rowanseymour:
All flow execution now happens in https://github.com/nyaruka/goflow which implements this feature

PR #940, opened 2019-06-06
parses variables in webhook header values (issue #770)

schuyler1d:
This is a bit of a drive by.  I was browsing issues in this project, but am unfamiliar with it -- however I am a little familiar with Django, so I believe this should implement #770 

However, I didn't see any tests that I could readily use and the test samples were a little too opaque for me.  Maybe someone else can create the test for this or someone could point me to an example that would help.  `test_webhook` looked the most promising, but I don't understand where the data comes from that is tested and grepping didn't help.


rowanseymour:
Hi @schuyler1d. We've actually moved flow execution to https://github.com/nyaruka/goflow which already implements this feature, and are in the process of ripping out the old engine code from this project. Sorry we neglected to close #770.

PR #955, opened 2019-06-14
Whatsapp Formatting

nicpottier:
We didn't have icons configured for whatsapp so for anon orgs it was especially confusing to know what channel a contact is communicating on. Also formats for non-anon orgs.


nicpottier:
wrong repo..

PR #992, opened 2019-07-03
🔥 populate account sid and and auth token on twilio callers when added

nicpottier:
Mailroom doesn't look at org configs, only channel configs (that way people can have multiple twilio accounts connected if they wanted in the future)

This sets the appropriate config options for callers when created off of the org settings.


nicpottier:
argh.. wrong repo

PR #1011, opened 2019-08-06
Fix `frame.haml` template

baracudda:
Indent the `brand.head_includes` and IE9 shim slightly so that when rendered into HTML, the code results inside the header rather than between the `<head>` and `<body>` tags.


nicpottier:
@ericnewcomer this look good to you?

PR #1171, opened 2020-03-11
Add prometheus monitoring

rudigiesler:
This abstracts the metrics portion of analytics.py into a generic metric provider, with provider implementations for Librato and Prometheus.

I've only added the concepts of a Gauge (something that goes up and down), and an Increment/counter (something that continually goes up), that is all that seems to be required at the moment.

I haven't abstracted out analytics part of analytics.py, since that's better suited for the current tools, rather than a metrics tool.

### DIFF
--- a/temba/utils/analytics/analytics.py
+++ b/temba/utils/analytics/analytics.py
@@ -5,17 +5,17 @@ from random import randint
 import analytics as segment_analytics
 from intercom.client import Client as IntercomClient
 from intercom.errors import ResourceNotFound
-from librato_bg import Client as LibratoClient
 
 from django.conf import settings
 from django.utils import timezone
 
 from temba.utils import json
 
+from .librato import LibratoBackend
+
 logger = logging.getLogger(__name__)
 
-# our librato_bg client
-_librato = None
+_metric_backends = []
 
 # our intercom client
 _intercom = None
@@ -44,14 +44,14 @@ def init_analytics():  # pragma: no cover
         _intercom = IntercomClient(personal_access_token=intercom_key)
 
     # configure Librato if configured
+    global _metric_backends
     librato_user = getattr(settings, "LIBRATO_USER", None)
     librato_token = getattr(settings, "LIBRATO_TOKEN", None)
     if librato_user and librato_token:
-        global _librato
-        _librato = LibratoClient(librato_user, librato_token)
+        _metric_backends.append(LibratoBackend(librato_user, librato_token))
 
 
-def get_intercom_user(email):
+def _get_intercom_user(email):
     try:
         return _intercom.users.find(email=email)
     except ResourceNotFound:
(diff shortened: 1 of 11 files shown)


temba/utils/analytics/analytics.py
```diff
@@ -60,47 +65,20 @@ def get_intercom_user(email):
...
-    reporting_hostname = "%s.%s" % (settings.MACHINE_HOSTNAME, settings.HOSTNAME)
+    global _metric_backends
+    for backend in _metric_backends:
+        backend.gauge(event, value)
 
-    if _librato:
-        _librato.gauge(event, value, reporting_hostname)
 
-
-def identify_org(org, attributes=None):
```
nicpottier: Is this being removed on purpose? We use intercom for support so we don't want that ripped out.

temba/utils/analytics/base.py
```diff
@@ -0,0 +1,17 @@
+import zope.interface
```
nicpottier: Doesn't Python3 have native interfaces that would work here?

temba/utils/analytics/prometheus.py
```diff
@@ -0,0 +1,53 @@
...
+            RELAYER_SYNC.labels(hostname=self._get_hostname()).set(value)
+        else:
+            logger.warning(f"Unknown prometheus gauge metric {{event}} with value {{value}}")
+
+    def increment(self, event, value=None):
+        if value is None:
+            value = 1
+
+        if event == "temba.contact_created":
+            CONTACTS.labels(hostname=self._get_hostname(), action="created").inc(value)
```
nicpottier: `created` label seems a bit redundant here, though maybe that's me no understanding prometheus properly.

temba/utils/analytics/analytics.py
rudigiesler: I didn't see anywhere in the code that was calling this, so I removed it. Happy to put it back if it is actually needed.

temba/utils/analytics/base.py
rudigiesler: Yeah I see that abstract base classes got an upgrade in 3.4 , they look much better now, let me try to use those and remove the dependancy.

temba/utils/analytics/prometheus.py
rudigiesler: This is more for if we want to at a later stage add updated/deleted actions, without having to change dashboard setups. I guess you can argue either way, right now it's not needed, the examples usually prefer having a single metric for the object, and having a label for each CRUD method.

Happy to do either method, since it doesn't matter at this moment with just a single action type.

nicpottier: Oh actually I missed that this was just recording a contacts counter, I thought it was using `contact_created` still. No, this seems great then actually.

One thing though is maybe these should still be prefixed with `temba_`? Grafana seems to build tree-ish browsing based on the underscore so it may make it a bit easier to look for rapidpro related events.

temba/utils/analytics/analytics.py
nicpottier: Huh, you may be right actually, this may all be happening in JS. @ericnewcomer can you confirm?

ericnewcomer: If memory serves, this was used as part of a migration and is no longer needed.

nicpottier:
This looks good to me, just waiting for 5.4 freeze to be over (likely Monday) before merging. We booted up a prometheus / grafana instance so are going to be dogfooding this too.

temba/utils/tests.py
```diff
@@ -1400,7 +1400,7 @@ def setUp(self):
         )
 
         self.intercom_mock = MagicMock()
-        temba.utils.analytics._intercom = self.intercom_mock
+        temba.utils.analytics.analytics._intercom = self.intercom_mock
```
rowanseymour: can move this test class to `temba/analytics/tests.py` now it has its own app

nicpottier:
One other piece here that occurred to me is that we probably want the `/metrics` endpoint gated with a username/password and basic auth no?

rudigiesler:
> One other piece here that occurred to me is that we probably want the `/metrics` endpoint gated with a username/password and basic auth no?

Yes, that is something we can do. 

A lot of setups use nginx to provide a basic auth restriction to the endpoint.

 I've done something slightly more fancy in the past, where we check if the request is going through the load balancer or not, and only allow requests that didn't go through the load balancer: https://github.com/praekeltfoundation/ndoh-hub/blob/develop/ndoh_hub/decorators.py#L6-L21

We can probably do a similar thing and wrap the view with basic auth, credentials in settings.

Thoughts on the best approach?

nicpottier:
Come to think of it I could probably be convinced that nginx settings are sufficient for this. I guess the main argument against it would be if we think we are leaking anything important in these metrics that we would feel some default installations would not lock down, in which case I would say defaulting to some basic authentication on the RapidPro config side would be better. But really we aren't talking about super sensitive information right now, # of contacts created etc isn't the end of the world.

rowanseymour:
@nicpottier @rudigiesler what's the status of this? Isn't this stuff that since ended up in mailroom?

nicpottier:
Not really, we did as prom endpoints to monitor a workspace, but this issue is more about global install monitoring.

rudigiesler:
Yeah sorry, never got the time to go back and complete this.

As Nic said, this is more of an "is the rapidpro instance happy/healthy" kind of thing, rather than getting stats about an org.

rowanseymour:
Closing this as we've since abstracted all of analytics in RapidPro

PR #1214, opened 2020-05-11
Add OrgActivity model that tracks contacts, active contacts, incoming and outgoing msgs

nicpottier:
Not used anywhere yet but want to start collecting it. Will be used to show activity on account page and growth of contacts etc..


temba/orgs/models.py
```diff
@@ -2724,3 +2725,76 @@ class BackupToken(SmartModel):
...
+    contact_count = models.IntegerField(null=True)
+
+    # the number of active contacts on this day
+    active_contact_count = models.IntegerField(null=True)
+
+    # the number of messages sent on this day
+    outgoing_count = models.IntegerField(null=True)
+
+    # the number of messages received on this day
+    incoming_count = models.IntegerField(null=True)
```
rowanseymour, line 2755: for simplicity could have these be non-null with zero defaults

nicpottier: Kinda went back and forth on this.. I think the reasoning I ended up with was that I'd rather know whether a value was written for that day rather than have it end up being zero just by default. From a storage perspective it is slightly cheaper as well for orgs that have no messages or active contacts on a day. I don't know, kinda of two minds about it honestly.

rowanseymour: I did too but then convinced myself that if there's a record for day X then those counts are all considered written... and this saves having the possibility of both zeros and NULLs in those columns

rowanseymour: I've also gone all go-brain these days so now I always try to avoid dealing with NULLs/nil unless the distinction between NULL and 0 is meaningful.. if we are going to create records with some counts calculated, some not.. then it's meaningful.. otherwise I don't think it is.  And ff we want to do any kind of arithmetic here it would be nice to not have to call COALESCE on every value.

PR #1250, opened 2020-06-24
🔥 Contact API endpoint can't call mailroom from within a transaction

rowanseymour:
I'm 99% this is the fix.. but still trying to repro the problem locally...

### DIFF
--- a/temba/api/v2/serializers.py
+++ b/temba/api/v2/serializers.py
@@ -656,6 +656,7 @@ class ContactWriteSerializer(WriteSerializer):
 
         if mods:
             self.instance.modify(self.context["user"], *mods)
+            self.instance.refresh_from_db()
 
         return self.instance
 
--- a/temba/api/v2/serializers.py
+++ b/temba/api/v2/serializers.py
@@ -7,6 +7,7 @@ import regex
 from rest_framework import serializers
 
 from django.conf import settings
+from django.db import transaction
 
 from temba import mailroom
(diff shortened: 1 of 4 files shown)


temba/api/v2/tests.py
```diff
@@ -1782,7 +1782,7 @@ def test_contacts(self, mr_mocks):
         self.assertEqual(jean.get_field_value(nickname), "Jado")
 
         # update by UUID and change all fields
-        with self.assertNumQueries(45):
+        with self.assertNumQueries(46):
```
rowanseymour, line 1785: extra hit for the refresh because we let mailroom make changes

temba/api/v2/serializers.py
```diff
@@ -630,30 +631,32 @@ def save(self):
...
-        # update our groups
-        if groups is not None:
-            mods += self.instance.update_static_groups(groups)
+            # update our groups
+            if groups is not None:
+                mods += self.instance.update_static_groups(groups)
 
         if mods:
             self.instance.modify(self.context["user"], *mods)
+            self.instance.refresh_from_db()
```
rowanseymour, line 659: separate issue - we send the contact back as the response so needs to be refreshed if mailroom made changes

nicpottier: This is going to be a race, we should probably be using the state that mailroom returns. (can address that separately)

Issue #1278, opened 2020-07-16
New WhatsApp BSP channel: 360dialog


johncordeiro:
hey team, today we've available in RapidPro 2 options to integrate with WhatsApp, the first one is the self-hosted WhatsApp installation and the other one Twilio WhatsApp. In order to give the flexibility for the users to choose among different fee rates, we have developed a great partnership with a German company called 360dialog.  They are a WhatsApp BSP and has a flat fee for integrating to WhatsApp, it sounds like a good alternative for the two available options.

By looking at their [API docs](https://docs.360dialog.com/api/wa-pipe-api/sandbox), it would be a new channel type but registered as a handler for the WhatsApp channel in courier.

nicpottier:
Thanks, ya, always good to have other WhatsApp options. What is their story on templates?

johncordeiro:
hey @nicpottier their templates API is not available yet, they're kind of creating a template API inspired on the FB one that we use, so it could be reused the WA channel templates as well. I'm contacting their support team to check.

jcbalmeida:
Hey guys, we opened the PR [#3083](https://github.com/nyaruka/rapidpro/pull/3083) to refresh and populate the message templates, since the template API is now available.

PR #1280, opened 2020-07-19
New WhatsApp BSP channel: 360dialog

johncordeiro:
I've created a new channel type as proposed in issue #1278. I've tested the approach that I quickly described by just creating a new channel type in RapidPro and registering this new channel type to the WhatsApp handler in courier and it works like a charm! Very straightforward integration.

You guys can test it by using the sandbox account, some instructions are available [at this page](https://docs.360dialog.com/api/whatsapp-api/sandbox). 

360Dialog still didn't make available the Template management API, but I'm in contact with their product team to follow a similar structure as the Messaging API, which is a proxy for the existing template management API from Facebook, so we can reuse a bunch of code that is created for the WhatsApp self-hosted channel as I've done so far with this channel.


matmsa27:
Hi everyone, do you have any news about this PR? Do you prefer that we open this PR in `nyaruka/rapidpro` ?

rowanseymour:
Replaced by https://github.com/nyaruka/rapidpro/pull/2980

PR #1283, opened 2020-07-20
🏁 Add start_type and created_by to queued flow starts

rowanseymour:
These can then be picked up by mailroom and added to the trigger as `user` and `origin`

### DIFF
--- a/temba/mailroom/queue.py
+++ b/temba/mailroom/queue.py
@@ -114,7 +114,9 @@ def queue_flow_start(start):
 
     task = {
         "start_id": start.id,
+        "start_type": start.start_type,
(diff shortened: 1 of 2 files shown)


temba/mailroom/queue.py
```diff
@@ -114,7 +114,9 @@ def queue_flow_start(start):
 
     task = {
         "start_id": start.id,
+        "start_type": start.start_type,
         "org_id": org_id,
+        "created_by": start.created_by.username,
```
rowanseymour, line 119: could be `created_by_id` but then mailroom has to load the user

rowanseymour:
wrong repo

Issue #300, opened 2016-07-27
RapidPro Over HTTPS


tipen:
Hi, I'm trying to setup RapidPro and hosting it on my test server but it runs HTTPS.

I'm getting the following errors,

```
[27/Jul/2016 06:12:59] code 400, message Bad request version ('\xc0\x13\xc0')
You're accessing the development server over HTTPS, but it only supports HTTP.
```

Is it supported? How can I enable SSL support?

Thanks!

nicpottier:
RapidPro is just a Django app, so the same settings and configuration for Django apply:
  https://docs.djangoproject.com/en/1.9/topics/security/

tipen:
Hi @nicpottier,

I tried changing SECURE_SSL_REDIRECT from False to True in env/lib/python2.7/site-packages/django/conf/global_settings.py file but I'm still getting error when accessing the GUI via https.

Any clue?

nicpottier:
Sorry, this is not in our scope to support, as I said, same as any other Django app.

PR #1296, opened 2020-08-04
Creation of new indexes to Channel model as solution of issue #300

Cloves23:
Solution to issue [300](https://github.com/nyaruka/mailroom/issues/300).


Cloves23:
PR to wrong repository.

PR #1311, opened 2020-08-19
Update frame.haml

JesusEkie:
Maintain consistency on the edit applied to settings_common.py on the brand.description field with the title displayed on the frontend

### DIFF
--- a/templates/frame.haml
+++ b/templates/frame.haml
@@ -5,7 +5,7 @@
   %head
     %title
       {% block page-title %}
-        {{ brand.name }} - Visually build interactive SMS Applications
+        {{ brand.name }} - {{ brand.description }}
       {% endblock %}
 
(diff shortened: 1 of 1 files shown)


rowanseymour:
Agree this can be customizable, tho I think maybe we want another setting in the brand that's a slightly shorter description than what we put on the splash page. Currently `brand.description` is `Visually build nationally scalable mobile applications from anywhere in the world.` and what we have in the title is `Visually build interactive SMS Applications`

Tho it's possible this has all been changed in the ongoing UI refresh @ericnewcomer ?

JesusEkie:
@rowanseymour I also think it will be a good idea to create a custom setting in the brand for that specific information. 
Can I suggest we add the setting "title" in the Branding Configuration?

rowanseymour:
Sounds ok.. @ericnewcomer ?

rowanseymour:
@JesusEkie we've now made title part of the brand so you can change it to whatever you want

PR #1384, opened 2020-11-18
Refresh Template messages for 360dialog Channels

jcbalmeida:
* Creates scheduled task (refresh_360_templates, each 15 minutes)
* Adds remaining templates for 360Dialog (templates.haml and sync_logs.haml)


jcbalmeida:
Ops, wrong base, changing to nyaruka/rapidpro

PR #1470, opened 2021-03-31
✨ Update translation files from transifex

rowanseymour:
I think I must have merged something that nuked a few translations - this resets us back to what is in transifex and gets rid of the djangojs.po/mo files we don't use anymore.


rowanseymour:
Looks like this PR was the culprit https://github.com/nyaruka/rapidpro/pull/3221

Issue #1481, opened 2021-04-22
Facebook page listing should be flexible


rafaelbreno:
In our platform there's some page administrators that manages a lot of pages, because of the limit of Page Listing(25 items) some users are reporting the need to increase the number and that they can't "uncheck" a page to "check" another(in the Facebook Popup), when doing this the "unchecked" page will stop working.

So, wouldn't be nice to add some another type of listing or increase the max number of pages?
![facebook_page_listing](https://user-images.githubusercontent.com/32229014/115760900-1fcf1480-a378-11eb-8c79-816803340a7b.png)

rafaelbreno:
A solution would be adding a local variable into rapidpro called `FACEBOOK_PAGES_LIMIT`, and it would be implemented as a field in the Facebook Graph url, there's this field called limit, so it would be something like:
`https://graph.facebook.com/me/accounts?limit={FACEBOOK_PAGES_LIMIT}`

matmsa27:
Hi @nicpottier and @rowanseymour , what do you think about this?

rowanseymour:
Can we just pass `limit=100`? What's the maximum we can request?

rafaelbreno:
Looking at the Facebook Graph API docs it does not show any limit, so in theory if my user administer 9999 pages I could fetch it using `https://graph.facebook.com/me/accounts?limit=9999` without any problem.

PR #1489, opened 2021-05-11
Feat/add fba pages limit

rafaelbreno:
Related to #1481 

Differences:
- Add `FACEBOOK_PAGES_LIMIT` environment variable into `settings_common.py`
- Add context to retrieve `FACEBOOK_PAGES_LIMIT` value
- Implement context value into _templates/channels/types/facebookapp/claim.haml_

### DIFF
--- a/temba/channels/types/facebookapp/views.py
+++ b/temba/channels/types/facebookapp/views.py
@@ -95,6 +95,7 @@ class ClaimView(ClaimViewMixin, SmartFormView):
         context = super().get_context_data(**kwargs)
         context["claim_url"] = reverse("channels.types.facebookapp.claim")
         context["facebook_app_id"] = settings.FACEBOOK_APPLICATION_ID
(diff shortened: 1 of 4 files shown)


rowanseymour:
This doesn't feel like it should be a setting... if we can't make a quick fix just by adding `limit=100` then I think we should be doing this properly by fetching all pages of results.

rafaelbreno:
@rowanseymour we've a account to manage the pages connections, so the total amount of manage pages vary, e.g at the moment this account manage approximately 137 pages, and with the time it can vary.

rowanseymour:
What are you planning to set the limit to? if it works for you to set it to `200` or whatever, there's no downside to using that number for everyone. It doesn't need to be a setting. If it doesn't work... then we need to do proper paging.

rafaelbreno:
#1495

Issue #1680, opened 2022-08-01
Performance issue in task `trim_http_logs_task`


Sandro-Meireles:
We are having problems with the `trim_http_logs_task` task, in our environment, this table has a large volume of data, 102.31 GiB to be exact. This causes the task to terminate without erasing any data, generating a "snowball effect". I discovered that the problem is precisely in the query, as it requires all data in the `trim_before` range to be returned.

I made an adjustment to the query that solves this problem inspired by the `trim_flow_sessions` task:

```py
...
    while True:
        http_log_ids = HTTPLog.objects.filter(created_on__lte=trim_before).values_list("id", flat=True)[:1000]

        if not http_log_ids:
            break

        HTTPLog.objects.filter(id__in=http_log_ids).delete()
        deleted_count += len(http_log_ids)

        logger.info(f"[ Trim HTTPLogs ] Deleted {deleted_count} HTTPLogs")
...
```
I would like to know if you have already noticed this problem and if this resolution makes sense.

rowanseymour:
We haven't seen that but your change looks reasonable - make a PR

PR #1685, opened 2022-08-18
Improve `trim_http_logs_task` performance by splitting the query

Sandro-Meireles:
Recently I did an investigation on this task, I found a performance problem in how the query was being made, after that I wrote this issue: https://github.com/rapidpro/rapidpro/issues/1680 
I was recommended to open a Pull Request with the suggested fixes, and here it is. I based myself on the `trim_flow_sessions` task because it is a very performant task.

### DIFF
--- a/temba/request_logs/tasks.py
+++ b/temba/request_logs/tasks.py
@@ -30,6 +30,6 @@ def trim_http_logs_task():
         num_deleted += len(http_log_ids)
 
         if num_deleted % 10000 == 0:  # pragma: no cover
-            print(f" > Deleted {num_deleted} http logs")
+            logger.debug(f" > Deleted {num_deleted} http logs")
 
     logger.info(f"Deleted {num_deleted} http logs in {timesince(start)}")
--- a/temba/request_logs/tasks.py
+++ b/temba/request_logs/tasks.py
@@ -1,15 +1,35 @@
+import logging
+
 from django.conf import settings
 from django.utils import timezone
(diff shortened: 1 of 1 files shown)


temba/request_logs/tasks.py
```diff
@@ -1,15 +1,35 @@
...
+        http_log_ids = HTTPLog.objects.filter(created_on__lte=trim_before).values_list("id", flat=True)[:1000]
+
+        if not http_log_ids:
+            break
+
+        HTTPLog.objects.filter(id__in=http_log_ids).delete()
+        num_deleted += len(http_log_ids)
+
+        if num_deleted % 10000 == 0:  # pragma: no cover
+            print(f" > Deleted {num_deleted} http logs")
```
rowanseymour, line 33: this should be a logger call - maybe `logger.debug`

rowanseymour, line 33: this should be a logger call - maybe `logger.debug`

rowanseymour (changes requested):

temba/request_logs/tasks.py
Sandro-Meireles: Thanks for the suggestion. I chose to use print based on an existing task [trim_flow_sessions](https://github.com/rapidpro/rapidpro/blob/main/temba/flows/tasks.py#L113). Maybe it makes sense to fix it too.

rowanseymour (approved):

PR #1703, opened 2022-11-01
fix: Improve `trim_channel_logs_task` performance by splitting the query

paulobernardoaf:
Recently we've had some problems with the performance on `trim_channel_logs_task` like in #1685.
This PR is to improve the task performance by splitting the query into batches replicating the behavior of `trim_flow_sessions` and `trim_http_logs_task`.


paulobernardoaf:
I have read the CLA Document and I hereby sign the CLA

PR #1745, opened 2025-01-21
Fix: Increase otp_secret length to 64 in orgs_usersettings

Wahesh:
The current length of otp_secret is 64. It throws an error while doing a new installation. Changing the length of otp_secret to 64 fixes the problem.


rowanseymour:
You must be using the wrong version of pyotp. The version pinned in poetry.lock uses secrets of length 16.

